PR #34, opened 2014-10-10
support for multiple velodynes

ddillenberger:
Hello,

we have a scenario here with multiple velodynes in one network. So we developed a patch for your driver, which allows working with multiple velodyne LIDARs.

It adds the parameter "device_ip" to the velodyne_driver node. With the parameter one can select which LIDAR's packets  should be processed and published by the node. Default behaviour is unchanged (process all packets).

Hope you find it useful.
Denis


trainman419:
Tried this on Hydro on 12.04, and it doesn't compile.

The error that I get is:

```
velodyne/velodyne_driver/include/velodyne_driver/input.h:84:5: error: ‘in_addr’ does not name a type
```

bricerebsamen:
`#include <netinet/in.h>` is missing in `input.h` to compile properly

PR #35, opened 2014-10-13
cut at specific angle feature

### LINKED ISSUE #33: Number of packets published per message is not exactly a revolution if "npackets" set to default
Hello@all,

we would like to submit a patch (velodyne_driver/driver.cc) which would cut the packet stream in a better way, so that a single revolution always starts/ends at a specific angle (for example while the LIDAR points to the robot).
There are two reasons for that:
First is that we've never seen a Velodyne LIDAR (we work with multiple 64E S2 and 32E models) whose rotation speed is very stable. Especially the  64E S2 models tend to be much slower than set when the enviroment temperature is low (10Hz measured instead of 15Hz set at 5-10°C). And our 32E models are varying between 9 and 11 Hz (indoor, same temperature), but all are set to 10Hz. Along with the varying rotation speed the number of packets per revolution vary a lot. This leads in the end to overlapping or missing sectors in the published pointclouds. (It actually looks quite cool in RViz :-D, but it can be a problem for some algorithms)
Second reason is that start and end of a revolution are an discontinuous area (if the robot moves), which can lead to wrong algorithm results in this area. With setting "the cut" to a specific angle you can move this area to an angle where it doesn't matter if you get wrong results.

Any objections?

Best regards
Denis

ddillenberger:
Hello,

this pull request should fix #33. I tested it sucessfully with a HDL 32E. We also used this code with an HDL 64E S2. The block-format never changed since 2008 I think. 
A new parameter "cut_angle" is introduced, where one can set the desired "cut angle" in degree in device coordinates.

Standard behaviour is unchanged, so no code should be broken by this fix.

Denis

### DIFF
--- a/velodyne_driver/src/driver/driver.cc
+++ b/velodyne_driver/src/driver/driver.cc
@@ -69,6 +69,10 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
   std::string dump_file;
   private_nh.param("pcap", dump_file, std::string(""));
 
+  double cut_angle;
+  private_nh.param("cut_angle", cut_angle, -0.01);
+  config_.cut_angle = int(cut_angle*100);
+
   // initialize diagnostics
   diagnostics_.setHardwareID(deviceName);
   const double diag_freq = packet_rate/config_.npackets;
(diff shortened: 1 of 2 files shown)


lemiant:
I've been using this PR in our project for a while. There is a bug where the timestamp is sometimes pulled from the wrong packet. Under this PR it still pulls the timestamp from 
````scan->packets[config_.npackets - 1].stamp````
even though the number of packets is no longer always npackets. This means that if the are more than npackets the timestamp is no longer from the last packet, which was the goal. Worse, when the numberof packets is less than npackets the timestamp is pulled from uninitialized memory off the end of the vector, which could be anything.
I've patched it for myself already, but I thought you might like to know.

**Comment by jack-oquin** [comment]
Thanks for the report @lemiant. 

I have merged this PR into a separate ``cut_at_specified_angle_feature`` branch. I would appreciate it if you would submit a new pull request against that for your fixes to it.

File: velodyne_driver/src/driver/driver.cc [hunk 1]
```diff
@@ -107,11 +111,45 @@ bool VelodyneDriver::poll(void)
...
+      {
+        int rc = input_->getPacket(&tmp_packet);
+        if (rc == 0) break;       // got a full packet?
+        if (rc < 0) return false; // end of file reached?
+      }
+      scan->packets.push_back(tmp_packet);
+
+      static int last_base_rotation = -1;
+      // Extract base rotation of first block in packet
+      std::size_t rot_data_pos = 100*0+2;
```
chukcha2: doesn't this always equal to 2? why have 100*0?

velodyne_driver/src/driver/driver.cc
```diff
@@ -107,11 +111,45 @@ bool VelodyneDriver::poll(void)
...
 
+  if( config_.cut_angle >= 0) //Cut at specific angle feature enabled
+  {
+    scan->packets.reserve(config_.npackets);
+    velodyne_msgs::VelodynePacket tmp_packet;
+    while(true)
+    {
+      while(true)
+      {
+        int rc = input_->getPacket(&tmp_packet);
```
volkandre: The API has changed here. getPacket now expects two arguments.
https://github.com/ros-drivers/velodyne/blob/3747e937cfc1b2986c0b2e4983bdd0d7e8934079/velodyne_driver/include/velodyne_driver/input.h#L80-L81

volkandre: This only leads to an error, if you try to merge this feature into the latest master branch. Please see https://github.com/ros-drivers/velodyne/pull/126 for an updated version of this feature.

JWhitleyAStuff:
Stale PR.  Updated with #126.

PR #36, opened 2014-10-17
Hdl32e imu gps receiver

ddillenberger:
Hello,

I implemented a IMU/GPS receiver for the HDL 32E LIDAR earlier this year.
It can be used together with the nmea_navsat_driver as it just extracts the NMEA sentence and publishes it. The IMU data are sent via a geometry_msgs/TwistStamped message.
The GPS clock is also extracted and sent via a std_msgs/Time message.

The node has two parameters "device_ip" and "udpport", so supports multiple devices.

Possibly you would  like to ship it with your driver.

Denis

### DIFF
--- a/velodyne_gps_imu/CMakeLists.txt
+++ b/velodyne_gps_imu/CMakeLists.txt
@@ -1,0 +1,165 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(velodyne_gps_imu)
+
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11") #or -std=c++0x
+
+## Find catkin macros and libraries
+## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
+## is used, also find other catkin packages
+find_package(catkin REQUIRED COMPONENTS
+  geometry_msgs
+  roscpp
+  std_msgs
+)
+
+## System dependencies are found with CMake's conventions
+# find_package(Boost REQUIRED COMPONENTS system)
+
+
+## Uncomment this if the package has a setup.py. This macro ensures
+## modules and global scripts declared therein get installed
+## See http://ros.org/doc/api/catkin/html/user_guide/setup_dot_py.html
+# catkin_python_setup()
+
+################################################
+## Declare ROS messages, services and actions ##
+################################################
+
+## To declare and build messages, services or actions from within this
+## package, follow these steps:
+## * Let MSG_DEP_SET be the set of packages whose message types you use in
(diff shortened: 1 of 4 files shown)


stuglaser:
The timestamp generation is very wrong.  Here are two echos from the time topic.

In this one, the seconds fail to wrap even though nanoseconds did:

```
---
data: 
  secs: 1432773818
  nsecs: 995018005
---
data: 
  secs: 1432773818
  nsecs: 998120070
---
data: 
  secs: 1432773818
  nsecs: 5062103
---
data: 
  secs: 1432773818
  nsecs: 7154942
```

Here, seconds wrap even though nanoseconds don't:

```
---
data: 
  secs: 1432773818
  nsecs: 554267883
---
data: 
  secs: 1432773818
  nsecs: 555624008
---
data: 
  secs: 1432773819
  nsecs: 564122915
---
data: 
  secs: 1432773819
  nsecs: 573288918
```

jpgr87:
The timestamp goofiness is the result of trying to merge times from two different sources: the velodyne's internal 1PPS conditioned clock, and the NMEA string used to set the clock reference to the 1PPS.  The velodyne doesn't get the NMEA string corresponding to a PPS pule for several hundred milliseconds after that pulse occurs due to the 4800bps serial link speed, so it is going to be inherently delayed in the data stream with respect to the timestamps the Velodyne is creating with its clock.  This transmission delay (plus whatever other processing delays the Velodyne introduces) is why the NMEA seconds appear only rolling over 500ms after the Velodyne's clock second rolls over.

The "time" message should really only be looking at the GPS information up to the current hour, and then adding the seconds information in the scan packet to offset the time since the last hour to the current message time.  That means you'll only have rollover mismatches once every hour instead.  So what you should *really* do is look at the GPS information once to determine the current hour, and increment it each time the scan packet's seconds field rolls over.

FWIW, this same timing method can also be used by the velodyne_pointcloud driver to correct the timestamps in the pointclouds to use the GPS timebase as discussed in issue 15.

jack-oquin:
I would love to have this included in the Velodyne driver. I like the way it is separately packaged.

To make it an official ROS driver:

 * The IMU should publish a standard ``sensor_msgs/Imu`` instead of ``geometry_msgs/TwistStamped``. Note that there are on-going discussions about that message, but it remains the ROS standard until a better one replaces it.
 * The different time bases and formats need to be addressed robustly.

jack-oquin:
Is this PR superseded by #53?

If so, let's close this one.

jpgr87:
I think so, 53 is an attempt to correct this pull request based on the review comments.

PR #38, opened 2015-01-08
VLP16

bricerebsamen:
this add support at the driver level for the new 16 beams velodyne model (Pucks, or VLP16)

### DIFF
--- a/velodyne_driver/src/driver/driver.cc
+++ b/velodyne_driver/src/driver/driver.cc
@@ -49,6 +49,10 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
     {
       packet_rate = 1808.0;
     }
+  else if (config_.model == "VLP16")    // generates ~300000 points per second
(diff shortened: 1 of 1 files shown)


bricerebsamen:
This has been running with success on our system for quite some time already. 

Note however that this PR only adds a trivial change to the driver. The real changes have to come in the velodyne_pointcloud files. At my place we have our custom code to convert from raw scan to point cloud.

The key is that the VLP-16 has only 16 beams but they are packet in a 32 beams packet. Which means that you get one encoder value for the 2 sets of 16 beams. In my implementation, I interpolate the encoder value for the second set as the mid-value between this packet's and next packet's encoder values.

jack-oquin:
Is this still relevant? There seem to be similar, overlapping updates. 

Please update your PR, if that seems worthwhile.

jack-oquin:
@bricerebsamen: I think this PR is no longer needed. Similar updates have already been merged. 

Should we close it, now?

PR #39, opened 2015-01-08
support for multiple velodynes (improved)

bricerebsamen:
This is an improvement on @ddillenberger implementation of the multi velodyne support. From what I understand from his design, and from what I experienced today on my setup (4 16-beam velodynes), @ddillenberger code allows to read data from one and only one laser at the same time.

With this pull request, we add support for reading data simultaneously from several lidars at the same time. Each lidar must be configured with a separate IP address and to broadcast on a separate port.

With this code, we were able to read from 4 velodyne at the same time.

Credits also go to @ddillenberger and @ahendrix for this pull request.

### DIFF
--- a/velodyne_driver/include/velodyne_driver/input.h
+++ b/velodyne_driver/include/velodyne_driver/input.h
@@ -34,6 +34,7 @@
 #include <unistd.h>
 #include <stdio.h>
 #include <pcap.h>
+#include <netinet/in.h>
 
 #include <ros/ros.h>
 #include <velodyne_msgs/VelodynePacket.h>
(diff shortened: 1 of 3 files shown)


bricerebsamen:
I forgot to mention: compared to @ddillenberger implementation, we added a parameter for the udp port to read from.

bricerebsamen:
This has been running with success on our system for quite some time already.

jack-oquin:
Good to know, Brice.

Sorry I have been so slow to work on this driver. I am not actively using it myself these days. 

Would you be interested in being a co-maintainer?

bricerebsamen:
OK. But I'm not using it much either. I have access to a few Pucks and a 64.
However, I'm using custom code for converting to a point cloud, because I'm using a different calibration file. So I would not be able to do much testing...

jack-oquin:
Is your calibration method proprietary, or would you be interested in sharing it?

bricerebsamen:
it's proprietary. 

The good news is that for the Pucks, it we don't need a calibration file.

bricerebsamen:
@jack-oquin OK I'll volunteer to be a maintainer

jack-oquin:
I appreciate your help.

PR #40, opened 2015-01-22
Add argument for velodyne port number.

gregjklein:
This allows multiple velodynes to broadcast to the same computer, each with a corresponding node to listen on the appropriate port.

### DIFF
--- a/velodyne_driver/src/driver/driver.cc
+++ b/velodyne_driver/src/driver/driver.cc
@@ -57,6 +57,9 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
   std::string deviceName("Velodyne HDL-" + config_.model);
 
   private_nh.param("rpm", config_.rpm, 600.0);
+  int udp_port = 0;
+  private_nh.param("port", udp_port, static_cast<int>(UDP_PORT_NUMBER));
+  config_.udp_port = udp_port;
   ROS_INFO_STREAM(deviceName << " rotating at " << config_.rpm << " RPM");
   double frequency = (config_.rpm / 60.0);     // expected Hz rate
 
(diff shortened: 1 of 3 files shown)


velodyne_driver/src/driver/driver.cc
```diff
@@ -92,7 +95,8 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
     }
   else
     {
-      input_.reset(new velodyne_driver::InputSocket(private_nh));
+      input_.reset(new velodyne_driver::InputSocket(private_nh,
+                                                    config_.udp_port));
```
jack-oquin: Why pass ``config_.udp_port`` explicitly? Isn't the whole ``config_`` structure already available?

gregjklein: So far as I can tell, the config_ structure isn't available inside of InputSocket, unless I were to pass it in. I figured it'd be easier to just pass the port number directly, so I wouldn't change the signature of InputSocket's constructor.

jack-oquin: Looks like you are right. 

Still, I wonder if it might be cleaner to have all the parameters available at that level.

gregjklein: It looks to me like the only field in config_ that is used in InputStream is the udp_port, so I'd be hesitant to add the entire struct.

jack-oquin: That's OK with me.

jack-oquin:
This no longer applies cleanly, probably because of someone else added a similar update.

Would you like to merge that and update this pull request?

jack-oquin:
Looks like this has already been merged from some other request. Closing now.

PR #41, opened 2015-01-26
fixed a calibration file parsing bug

prclibo:


jack-oquin:
Thanks for the bug fix!

vrabaud:
actually, this makes some test fail silenty:
```
ctest -R tests_cloud_node_hz -V
```
8: YAML Exception: yaml-cpp: error at line 0, column 0: bad conversion
8: [ERROR] [1430825961.083218792]: Unable to open calibration file: /home/vrabaud/workspace/ws_velodyne/src/velodyne/velodyne_pointcloud/params/64e_utexas.yaml

Which is normal as ```horiz_offset_correction``` is not part of https://github.com/ros-drivers/velodyne/blob/master/velodyne_pointcloud/params/64e_utexas.yaml

What should be done ? Can we re-generate those files in the params folder ? Should we assume that if ```horiz_offset_correction``` is not present, it is the same as ```vert_offset_correction``` ?

PR #42, opened 2015-01-26
fixed gen_calibration min/max intensity type

prclibo:
`min_intensity` and `max_intensity` should be saved as integer, otherwise [this line](https://github.com/ros-drivers/velodyne/blob/master/velodyne_pointcloud/src/lib/calibration.cc#L72) will break


jack-oquin:
Thanks for the patch!

Issue #24, opened 2014-03-28
Proposed addition of GPS and IMU data


ibaranov-cp:
Hey,

I was wondering why no one ever added in the GPS and IMU data.
I am going to be working on this, and might ask for a merger later.

SISegwayRmp:
I'll put up an IMU driver for the HDL-32E soon, we used it to gather IMU data for online 3D mapping but gave up the 32 for a few 16's. I have quite a bit of inertial estimation experience and I must warn that the 32E inertial sensors are not temperature or bias corrected; they are good enough for gross correction since they are frame aligned with laser data and the accs are redundant. 

More useful is the GPS NTPD server I used that aligned the laser firing data with our odometry data on the embedded side, it eliminates ghosts at high speed (>6m/s).  But since it is so tied into our embedded state estimate it'll probably be a while before I can come up with some way include that. I will include it with the Segway RMP V3 stack that we'll be releasing soon.

jack-oquin:
Sounds great, thanks!

ksatyaki:
Is there any update on this? Has the GPS been interfaced?

ibaranov-cp:
We ended going with an external, higher quality data source anyway.
The onboard IMU and GPS re-boradcast is not great.
Possibly this is being eliminated in the newer small "puck" velodynes anyway.

PatrickHussey:
Agreed we ended up standardizing on the VLP-16 which does not have the IMU data, we usually use a MicroStrain 3DM-GX3-35 (or 45) in conjunction with the Velodyne sensors. 

Like ibaranov-cp said, the Velodyne inertial sensors are not calibrated and the GPS re-broadcast is pretty limited. Go with a good calibrated inertial sensor and a GPS that doesn't get relayed by the Velodyne.

va9abund:
Hello,

I'd like to ask you for hint, on the branch gps_imu there's interesting node velodyne_gps_imu, however there ain't any launch file inside. And rqt_graph shows nothing from pub/subscribers used in source files. Has anyone have done any tests of this? 

I've got VLP-16 and I'd like to improve SLAM-mapping using position from GPS.
As you can see from the image when I go and back it loses odometry and I end up with 2 different clouds. I will appreciate any help. 

![capture](https://user-images.githubusercontent.com/33001302/36140468-aa0009ec-10a1-11e8-9922-c3cace0bf934.PNG)

JWhitleyWork:
[...]

PR #44, opened 2015-03-30
adding driver and pointcloud support for the VLP16

SISegwayRmp:
Modifications to support the VLP16 including pointcloud processing.

I have the a PCAP file if there is no hardware available to test. I have tested with the units we have inhouse. Also added azimuth correction for inter block/firing dsr#.

Added launch file and calibration for the VLP16 as well.

### DIFF
--- a/velodyne_driver/CMakeLists.txt
+++ b/velodyne_driver/CMakeLists.txt
@@ -54,12 +54,19 @@ if (CATKIN_ENABLE_TESTING)
     http://download.ros.org/data/velodyne/32e.pcap
     DESTINATION ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_SHARE_DESTINATION}/tests
     MD5 e41d02aac34f0967c03a5597e1d554a9)
+  catkin_download_test_data(
+    ${PROJECT_NAME}_tests_vlp16.pcap
+    http://download.ros.org/data/velodyne/vlp16.pcap
+    DESTINATION ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_SHARE_DESTINATION}/tests
+    MD5 f45c2bb1d7ee358274e423ea3b66fd73)
   
   # unit tests
   add_rostest(tests/pcap_node_hertz.test)
   add_rostest(tests/pcap_nodelet_hertz.test)
   add_rostest(tests/pcap_32e_node_hertz.test)
   add_rostest(tests/pcap_32e_nodelet_hertz.test)
+  add_rostest(tests/pcap_vlp16_node_hertz.test)
+  add_rostest(tests/pcap_vlp16_nodelet_hertz.test)
   
   # parse check all the launch/*.launch files
   roslaunch_add_file_check(launch)
--- a/velodyne_driver/src/driver/driver.cc
+++ b/velodyne_driver/src/driver/driver.cc
@@ -36,25 +36,34 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
   // get model name, validate string, determine packet rate
   private_nh.param("model", config_.model, std::string("64E"));
   double packet_rate;                   // packet frequency (Hz)
+  std::string model_full_name;
   if ((config_.model == "64E_S2") || 
       (config_.model == "64E_S2.1"))    // generates 1333312 points per second
     {                                   // 1 packet holds 384 points
       packet_rate = 3472.17;            // 1333312 / 384
+      model_full_name = std::string("HDL-") + config_.model;
     }
   else if (config_.model == "64E")
     {
       packet_rate = 2600.0;
+      model_full_name = std::string("HDL-") + config_.model;
     }
   else if (config_.model == "32E")
     {
       packet_rate = 1808.0;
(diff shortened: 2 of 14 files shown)


jack-oquin:
That looks great! Thanks for the contribution.

We need a PCAP file to use for a unit test case. Preferably no more than 20MB or 30MB. It'll be uploaded to download.ros.org for the build farm to use.

SISegwayRmp:
No problem, big fans of Velodyne here. We just got 4 VLP-16. I'll upload the PCAP soon. 

I also have a driver for the HDL-32E position stream if that's of interest. No good to me now; the VLP-16 doesn't have inertial sensors (looks like I'm back to the Segway BSA or a 3DM-GX3-45 inertial reference for 3D mapping). One note is that the inertial sensors on the Velodyne systems don't appear to be temperature compensated so not really a big loss, plus the GPS is only available in RMC which is no good for covariance or 3D fix. Although GPS is somewhat useful for precise time synchronization using a NTPD driver.

jack-oquin:
> I also have a driver for the HDL-32E position stream if that's of interest.

There have been many requests for that. Does it run as a separate node subscribing to a different UDP port?

SISegwayRmp:
Yes it is a completely separate node running in Python with a separate UDP connection, I'll include it in the velodyne driver and create a new pull request when I have a chance.

I have a PCAP file ready, where should I upload it? I also updated the test directories to include the VLP16 tests, do you want me to add those to the pull request? They are all correct except for the MD5 checksum of the file on download.ros.org, which I won't have until the file gets uploaded.

Also I noticed in the PCAP files on download.ros.org do not have messages directed to ports other than 2368 filtered out which should cause the RawData::unpack function to fail, but instead because the headers are not specifically checked in the 32E/64 HDL unpack function it causes this strange blind spot behavior when playing back PCAP files, because those packets are skipped but not decremented from the number of packets per rotation. At any rate the PCAP file I made filters by port and the RawData::unpack_vlp16 has an assertion check for the header (0xEEFF), and no blind spots in the PCAP playback.

jack-oquin:
> Yes it is a completely separate node running in Python with a separate UDP connection, I'll include it in the velodyne driver and create a new pull request when I have a chance.

How about adding a new package to the repo, called ``velodyne_position``, or something similar? It's only useful for certain models. 

Alternatively, if it's easier to add as a new node to ``velodyne_driver`` (because of common code or something), that's OK with me. The main benefit of a separate package is that it does not add additional dependencies for those not using the new node.

> I have a PCAP file ready, where should I upload it? I also updated the test directories to include the VLP16 tests, do you want me to add those to the pull request? They are all correct except for the MD5 checksum of the file on download.ros.org, which I won't have until the file gets uploaded.

It needs to go in ``http://download.ros.org/data/velodyne/``, with a name like ``vlp16.pcap``.

jack-oquin:
You can send the PCAP directly to @tfoote, or send it to me and I'll pass it to him. It might be possible to send as an e-mail attachment. How large is the file?

Let's not add the test until the upload finishes. Otherwise, the build farm's continuous integration will start reporting failures.

The MD5 sum should only depend on the file's contents, not its location.

SISegwayRmp:
got it I'll send you a link to dropbox, its 9MB so email will probably just kick it back

SISegwayRmp:
Here is a link to the pcap file

https://www.dropbox.com/s/rljaapjn9rzx0id/vlp16.pcap?dl=0

tfoote:
The file is now available at: http://download.ros.org/data/velodyne/vlp16.pcap

jack-oquin:
Thanks, Tully.

SISegwayRmp:
I committed the updates to include the tests for VLP16, update launch header and modified the CMakeLists.txt for velodyne_driver and velodyne_pointcloud to include the tests and download the test data.

When I get some time I'll create a new package for the position data parsing and time synchronization with GPS via NTPD. We actually use GPS and GPS denied RTC time synchronization with our embedded system so I might just release it as supported for our platforms when we release the stack for the new Segway system architecture we developed.

jack-oquin:
Sounds good. There is already an issue (#24) requesting it.

SISegwayRmp:
Ran all the tests relating to the VLP16 support, should be good to go.

jack-oquin:
Works for me. The only test problems are due to #45, and have nothing to do with this PR.

Thanks for the excellent work. Whom should we credit in the change history?

SISegwayRmp:
Patrick Hussey, haven’t gotten around to a personal github account yet.

PATRICK HUSSEY
Principal Engineer

STANLEY INNOVATION, INC
9 Henry Clay Drive
Merrimack, NH 03054
Direct: 603.689.9724
Fax: 603.882.8818
http://www.stanleyinnovation.com<http://www.stanleyinnovation.com/>

jack-oquin:
No problem, Patrick. Thanks for your contribution.

SISegwayRmp:
Happy to help, I think the VLP16 is a game changer. Our platforms push 8m/s+ so the Velodyne systems are perfect for us. We plan to really push support of the sensor going forward. I am very impressed with the sensor, and Velodyne LiDAR products in general.

Glad I could contribute!

PR #46, opened 2015-04-13
Added filter to ignore position data while reading from pcap file

jitrc:
While playing pcap file of Velodyne 32E I found gaps in the scan which down not occur when directly connecting to hardware. 
I happens so as when connecting to hardware it connects to LIDAR data port 2368 and ignores position port 8308. But while reading pcap file it reads all data packets treating position data as LIDAR data which is not parsed correctly and causes a gap in the point cloud data.

Currently just ignored those packets, later would like to publish those packet in a separate topic both from port 8308 when connected to hardware and from pcap file so that GPS data can be utilized.

### DIFF
--- a/velodyne_driver/src/lib/input.cc
+++ b/velodyne_driver/src/lib/input.cc
@@ -240,10 +240,22 @@ namespace velodyne_driver
             // Keep the reader from blowing through the file.
             if (read_fast_ == false)
               packet_rate_.sleep();
-            
-            memcpy(&pkt->data[0], pkt_data+42, packet_size);
-            pkt->stamp = ros::Time::now();
-            empty_ = false;
+            if((header)->len - 42 == 1206)
+            {
+				memcpy(&pkt->data[0], pkt_data+42, packet_size);
+				pkt->stamp = ros::Time::now();
+				
+			}else if((header)->len-42 == 512)
+			{
+				ROS_INFO_ONCE("Position data ignored ");
+                return 1;
+			}else
+			{
+				ROS_WARN("Incorrect data length %d (1206 LIDAR data, 512 position data): ",
+                     (header)->len-42);
+                return 1;
+			}
+			empty_ = false;
             return 0;                   // success
           }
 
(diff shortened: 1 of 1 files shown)


jack-oquin:
Thanks for the fix. Would it make sense to check the port number rather than packet length?

It is fine to convert tabs to spaces, but please follow the existing "emacs-gnu" indentation style. Consistency makes the code easier to read and maintain, while unnecessary re-indenting obscures the diff's.

jitrc:
I thought of that too but din't find a easy way to do it while briefly looking into pcap functions.
also the header data does not have much more info 
```cpp
struct pcap_pkthdr {
	struct timeval ts;	/* time stamp */
	bpf_u_int32 caplen;	/* length of portion present */
	bpf_u_int32 len;	/* length this packet (off wire) */
};
```
Surely the port information is stored some where as I can see it while opening in wireshark, but instead of looking any further for a solution I decided to look into velodyne's official software, VeloView (http://www.paraview.org/Wiki/VeloView).
There only I found that they are using data length to decide weather a package has lidar data or position data irrespective of velodyne model while it uses port no. while connecting to the hardware directly.
refer: https://github.com/Kitware/VeloView/blob/master/VelodyneHDL/PacketFileSender.cxx
```cpp
if (!this->Internal->PacketReader->NextPacket(data, dataLength, timeSinceStart))
    {
    this->Internal->Done = true;
    return;
    }

  // Recurse until we get to the right kind of packet
  if (dataLength == 1206)
    {
    ++this->Internal->PacketCount;
    size_t bytesSent = this->Internal->LIDARSocket->send_to(boost::asio::buffer(data, dataLength),
                                                            this->Internal->LIDAREndpoint);
    }

  if( (dataLength == 512) )
    {
    size_t bytesSent = this->Internal->PositionSocket->send_to(boost::asio::buffer(data, dataLength),
                                                               this->Internal->PositionEndpoint);
    }
}
```

jack-oquin:
OK. Might be worth adding a comment to that effect.

jack-oquin:
I have come up with a different solution to this problem. 

Please check out the ``pcap_port`` branch, and let me know whether it solves your problem.

jack-oquin:
The alternative fix for this issue has now been merged into the master branch. Since there have been no test results posted, I am closing this issue. 

If tests *are* run and any problem is observed, please re-open. Thanks!

Issue #45, opened 2015-04-02
fail to read calibration


bricerebsamen:
When running the tests, it complains about not being able to read the calibration file:

```
[velodyne_pointcloud.rosunit-cloud_nodelet_32e_hz_test/test_hz][passed]

SUMMARY
 * RESULT: SUCCESS
 * TESTS: 1
 * ERRORS: 0
 * FAILURES: 0

rostest log file is in /home/default/.ros/log/rostest-zoox164-28207.log
-- run_tests.py: verify result "/home/default/zoox_ros_ws/build/test_results/velodyne_pointcloud/rostest-tests_cloud_nodelet_32e_hz.xml"
[100%] Built target _run_tests_velodyne_pointcloud_rostest_tests_cloud_nodelet_32e_hz.test
-- run_tests.py: execute commands
  /opt/ros/indigo/share/rostest/cmake/../../../bin/rostest --pkgdir=/home/default/zoox_ros_ws/src/velodyne/velodyne_pointcloud --package=velodyne_pointcloud --results-filename tests_cloud_node_hz.xml /home/default/zoox_ros_ws/src/velodyne/velodyne_pointcl
oud/tests/cloud_node_hz.test
... logging to /home/default/.ros/log/rostest-zoox164-28616.log
[ROSUNIT] Outputting test results to /home/default/zoox_ros_ws/build/test_results/velodyne_pointcloud/rostest-tests_cloud_node_hz.xml
YAML Exception: yaml-cpp: error at line 0, column 0: bad conversion
[ERROR] [1427931670.442163378]: Unable to open calibration file: /home/default/zoox_ros_ws/src/velodyne/velodyne_pointcloud/params/64e_utexas.yaml
testcloud_node_32e_hz_test ... ok
```

Note the `ERROR` towards the end. In gdb I traced it to calibration.cc:70 where it reads `HORIZ_OFFSET_CORRECTION`. The corresponding field does not exist in the calibration file. Was probably introduced in fc20169be77cb7f649961fb97f91e565fe9ba28a

jack-oquin:
Yes. I've seen this recently, too. That field did not exist in the old 64E model.

I am not sure what to do. Do you have a suggestion? Should we make the field optional, or change the default calibration to some newer model?

bricerebsamen:
I think it depends on how critical this parameter is, i.e. can it be optional and/or have a default value. It also depends whether you want to invalidate all the calibration files that people have generated, more or less painstakingly, over the years.

[...]

PR #49, opened 2015-05-07
remove PCL dependency

vrabaud:
this is just a proof of concept at this point to see if you are interested in that: to make it fully work, we would have to add the `point_cloud2_iterator.h` from sensor_msgs that are only in Indigo and above. Or we could branch for Hydro. Or wait for Hydro to be eol-ed.

Some tests are currently failing but that's because of #45

### DIFF
--- a/velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
@@ -19,28 +19,39 @@
 #ifndef __VELODYNE_POINTCLOUD_POINT_TYPES_H
 #define __VELODYNE_POINTCLOUD_POINT_TYPES_H
 
-#include <pcl/point_types.h>
+#include <sensor_msgs/PointCloud2.h>
+#include <sensor_msgs/point_cloud2_iterator.h>
 
 namespace velodyne_pointcloud
 {
   /** Euclidean Velodyne coordinate, including intensity and ring number. */
-  struct PointXYZIR
+/*  struct PointXYZIR
   {
     PCL_ADD_POINT4D;                    // quad-word XYZ
     float    intensity;                 ///< laser intensity reading
     uint16_t ring;                      ///< laser ring number
     EIGEN_MAKE_ALIGNED_OPERATOR_NEW     // ensure proper alignment
   } EIGEN_ALIGN16;
+*/
 
+  inline
+  void setVelodyneCloudFields(sensor_msgs::PointCloud2 &pc) {
+    sensor_msgs::PointCloud2Modifier modifier(pc);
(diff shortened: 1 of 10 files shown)


vrabaud:
rebased, all tests passing. FYI @jacquelinekay

velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
```diff
@@ -19,28 +19,39 @@
...
 
-#include <pcl/point_types.h>
+#include <sensor_msgs/PointCloud2.h>
+#include <sensor_msgs/point_cloud2_iterator.h>
 
 namespace velodyne_pointcloud
 {
   /** Euclidean Velodyne coordinate, including intensity and ring number. */
-  struct PointXYZIR
+/*  struct PointXYZIR
```
jacquelinekay: Why keep this commented struct?

jack-oquin: I think it's actually needed by various packages.

vrabaud: Do we know which packages ? I cannot tell from http://rosindex.github.io/p/velodyne_pointcloud/github-ros-drivers-velodyne/#indigo-deps (there should be some PR2 stuff in there no ?)

The packing of that struct is using PCL which means that to read it, you also need PCL. Maybe a break of compatibility using that PR would help in making that format be easier to read/write.

jack-oquin: The driver defines a custom point type, so it can publish the laser ring number with the X, Y, Z and intensity of each point.

Getting rid of that is definitely an API breakage.

vrabaud: right it is, and as mentioned before, relying on PCL for encoding means you'll need it for de-encoding. Plus the packing usually contains unused bit for memory alignment.
And Kinetic is coming so it could be a good time too :) Or at least, please close this PR if you deem it inappropriate.

jack-oquin: Eliminating the PCL dependency would be nice.

I'll think some more about how to handle the API breakage. Maybe doing that in Kinetic would be allowable, I suppose. The trouble is: most users are still on Indigo, and many of them build it from source.

jacquelinekay: If you're going to break API, doing it for Kinetic would be the best time.

jack-oquin:
After considering this issue, my personal reaction is that it's not a good idea to change the API for Kinetic, but it could be *very* worthwhile for ROS 2, which currently has no PCL support.

What do y'all think about that idea?

JWhitleyAStuff:
I agree with Jack on this one. This PR is also incredibly stale.

PR #51, opened 2015-06-22
Added vertical sin angle correction

kunlileo:
So here is the full rawdata.cc file with the corrected codes as I suggested in the Issue#50. Like I explained there, I changed some of the expressions for the calculations of x, y, z and we got in our Tests a better result, which is show down here:
the result with the original codes:
foto 1
the result with the new codes:
foto 3
 the result with the original codes:
foto 2 1
the result with the new codes:
foto 1 1
 the result with the original codes:
foto 2
 the result with the new codes:
foto 4


jack-oquin:
Thanks for providing this fix and testing it.

PR #55, opened 2015-07-21
Fixed azimuth overflow bug.

lemiant:
Fix for VLP16 unpack:
For interpolated azimuth values between 35999.5 and 36000.0 the nested round(fmod())
yields a value of 36000 which is invalid and overflows the pre-computed sin/cos arrays,
since they only go form 0..35999
I switched the order of rounding and modulo so this doesn't happen.


jack-oquin:
Good catch. Thanks!

PR #67, opened 2015-11-17
Fix velodyne

pdsljp:
Move launch files in velodyne package into velodyne_pointcloud/launch ,because velodyne package is meta package

Change driver.cc in velodyne_driver package in order to use HDL-64E S3, but packet rate is experimental value


pdsljp:
Sorry. This is mistake.

jack-oquin:
Thanks for your interest. We should probably add 64E S3 support.

How does it differ from the 64E S2?

velodyne_driver/src/driver/driver.cc
```diff
@@ -43,6 +43,11 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
       packet_rate = 3472.17;            // 1333312 / 384
       model_full_name = std::string("HDL-") + config_.model;
     }
+ 　else if (config_.model == "64E_S3")
+    {
+      packet_rate = 5800.0;　// experimental
```
syn1993: Hi, I know it has been a long time since the last conversation, but I want to know whether the value of the packet_rate(5800) is right for HDL64E_S3? Because I'm getting the same question that I get around 16.5 hz on the /velodyne_packets topic when the packet_rate is set to 3472.17 and 10.1 hz when 5800. So, does it mean the packet_rate (5800) is right? However, I found you said it's a mistake.To be honest, I am confused here.

wawaafan:
You can refer to velodyne instruction manual, download address http://www.velodynelidar.com/downloads.html

PR #68, opened 2015-11-29
Remove unused variable

jlblancoc:
I think that `dsr` was unused. See line 317: 

for (int dsr=0; ...


jack-oquin:
This whole package needs a bit of clean-up. Thanks.

PR #73, opened 2016-02-09
Correct important data type error for VLP-16

fudger:
The timing constants for theVLP-16 azimuth calculation are of type `int` instead of `float` or `double`. Therefore, the computed point cloud is distorted.
Apart from that, I rearranged a few comments. Feel free to include these latter changes or not.


jack-oquin:
Looks good. Thanks!

PR #74, opened 2016-02-09
Add method to organize point clouds for VLP-16

acschaefer:
Add a method that organizes a 1D point cloud read from VLP-16. As @spuetz pointed out, this may accelerate post-processing steps. 

I tried to alter as little code as possible.

I tested the code using multiple VLP-16. Other model's data streams will still be processed as 1D clouds

### DIFF
--- a/velodyne_pointcloud/src/lib/rawdata.cc
+++ b/velodyne_pointcloud/src/lib/rawdata.cc
@@ -158,6 +158,8 @@ namespace velodyne_rawdata
              ||(config_.min_angle > config_.max_angle 
              && (raw->blocks[i].rotation <= config_.max_angle 
              || raw->blocks[i].rotation >= config_.min_angle))){
+          
+          // convert polar coordinates to Euclidean XYZ
           float distance = tmp.uint * DISTANCE_RESOLUTION;
           distance += corrections.dist_correction;
   
@@ -258,7 +260,7 @@ namespace velodyne_rawdata
   
           if (pointInRange(distance)) {
   
-            // convert polar coordinates to Euclidean XYZ
+            // append this point to the cloud
             VPoint point;
             point.ring = corrections.laser_ring;
             point.x = x_coord;
@@ -266,7 +268,6 @@ namespace velodyne_rawdata
             point.z = z_coord;
             point.intensity = (uint8_t) intensity;
   
-            // append this point to the cloud
             pc.points.push_back(point);
             ++pc.width;
           }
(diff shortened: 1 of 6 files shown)


spuetz:
In what points the code differs from mine in rawdata.cc? (https://github.com/spuetz/velodyne/commit/7c4950fcf34f613ddd0e01e0001f72995c8e3a5c) Don't you have to sort each 2D scan line? Or how the VLP-scan is structured? It would be nice, if we can create one general solution together.

velodyne_pointcloud/src/lib/rawdata.cc
```diff
@@ -435,19 +436,27 @@ namespace velodyne_rawdata
...
+            else {
+              
+              // If not in the specified range, set point to NaN.
+              point.x = std::numeric_limits<float>::infinity();
+              point.y = std::numeric_limits<float>::infinity();
+              point.z = std::numeric_limits<float>::infinity();
+              point.intensity = 0u;
+            }
+
+            pc.points.push_back(point);
```
spuetz: Don't you have to sort each 2D scan line? Or how the VLP-16 scan is structured?

fudger: We take different approaches to organize the point clouds.
- You organize the points in RawData::unpack() and you do not provide support for the VLP-16.
- I hardly change RawData::unpack(), but I take the 1D point cloud it produces and organize it in [Convert](https://github.com/fudger/velodyne/blob/ordered_point_cloud_vlp16/velodyne_pointcloud/src/conversions/convert.cc#L84). My code only supports the VLP-16.

I would be happy if we could devise a general solution together! Perhaps we can build upon [organize_point_cloud](https://github.com/fudger/velodyne/blob/organize_point_cloud/velodyne_pointcloud/src/lib/rawdata.cc), a branch I started to organize point clouds coming from any Velodyne sensor? The clue in this branch is that RawData::unpack() takes Velodyne messages instead of individual packets.

PS: The idea of processing whole messages instead of packets is not only suited to organize point cloud. It comes in handy when [accurately transforming points](https://github.com/fudger/velodyne/blob/per_point_transform/velodyne_pointcloud/src/lib/rawdata.cc), too.

fceex49:
Actually I have noticed two issues in this PR
- the is_dense flag is not filled correctly
- also the invalid points should be marked as NaN (otherwise many things in PCL or rospy will not work)

JWhitleyAStuff:
@spuetz / @acschaefer - Any movement on this feature? Suggestions for making a general version?

acschaefer:
I am sorry! Currently, I am not able to contribute to the development of this feature, as I am busy with other projects. Organized point clouds would be a very useful feature, though.

JWhitleyAStuff:
Thanks for the response, @acschaefer. @versatran01 said he may be committing a new version of this. I'm going to close this PR for the time being.

PR #75, opened 2016-02-12
Support for VLP-16's dual-return mode

fudger:
Added a few changes to automatically adapt the way point clouds are calculated to the current return mode of the VLP-16: strongest, last, or dual.

The dual-return mode, which is not yet supported in the current version of velodyne_pointcloud, is of great use not only in forestry applications; if offers more data when looking through transparent objects in standard mapping applications.

In detail, the following functionality is added:
1. Read the factory bytes in the Velodyne packet to detect in which return mode the sensor is running.
2. Adapt the azimuth calculation according to the return mode.
3. In dual return mode, output one point cloud for two received packets. In this way, every scan represents a full scanner revolution.

### DIFF
--- a/velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
@@ -166,7 +166,21 @@ namespace velodyne_rawdata
     float cos_rot_table_[ROTATION_MAX_UNITS];
     
     /** add private function to handle the VLP16 **/ 
+    /** @brief convert raw VLP16 packet to point cloud
+     *
+     *  @param pkt raw packet to unpack
+     *  @param pc shared pointer to point cloud (points are appended)
+     */
     void unpack_vlp16(const velodyne_msgs::VelodynePacket &pkt, VPointCloud &pc);
+    
+    /// \brief Read one VLP-16 firing sequence (half a block).
+    /// \param[in] data pointer to first range-intensity data element.
+    /// \param[in] azimuth azimuth angle of first data element.
+    /// \param[in] azimuth_diff azimuth difference to first element of next firing.
+    /// \param[out] pc point cloud to which the read points are appended.
+    void read_firing_vlp16(const uint8_t* data, 
+                           float azimuth, float azimuth_diff,
+                           VPointCloud &pc);
 
     /** in-line test whether a point is in range */
     bool pointInRange(float range)
--- a/velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
@@ -66,6 +66,11 @@ namespace velodyne_rawdata
     Dual      = 0x39
   };
   
+  enum SensorModel {
+    Hdl32e  = 0x21,
+    Vlp16   = 0x22
+  };
+  
   /** Special Defines for VLP16 support **/
   static const int    VLP16_FIRINGS_PER_BLOCK =   2;
   static const int    VLP16_SCANS_PER_FIRING  =  16;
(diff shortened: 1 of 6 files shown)


jack-oquin:
I don't think those factory bytes are present for the various HDL-64 models. 

Is that being handled, somehow?

fudger:
Hi Jack, thanks for your response!

You were right: after perusing the HDL64-E manual, I found out that the HDL64-E uses the same factory byte values for publishing different kinds of information. So now the model is identified by the number of lasers specified in the calibration.

velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
```diff
@@ -138,11 +187,24 @@ namespace velodyne_rawdata
      */
     int setup(ros::NodeHandle private_nh);
 
+    /** @brief convert raw packet to point cloud
+     *
+     *  @param pkt raw packet to unpack
+     *  @param pc shared pointer to point cloud (points are appended)
+     *  @param[out] m return mode of the laser scanner: last, strongest, or dual
```
jlblancoc: `m` seems to be spurious here, right??

fudger: Thanks! Your hint will find its way into my next commit. Already working on it.

altrouge:
Hi, what is the current state of this feature ? I am also interested in vlp16 dual returns.

From what I understand, you merge 2 consecutive scans to get a full revolution. I think a cleaner way would be to double the packet_rate in the driver, though I have no idea how to do it without adding a new parameter.
I think it would also be interesting to have the possibility to separate the pointclouds afterwards (last and strongest), for instance, to have algorithms using the structure of the pointcloud in strongest/last mode. If you want, I can help with that.

fudger:
Hi, I am glad you are interested in this branch. To answer your questions:

### Current state
The branch is tested with multiple VLP-16 scanners in our lab and is running without errors. The visualization with rviz works fine.

### Working principle
You are right, I am merging two consecutive scans to get a full revolution. To determine whether I have to read one or two packets to get a full revolution, I use the factory bytes in the Velodyne scan message. They encode the mode of operation.

### How to distinguish between different kinds of returns
With little postprocessing, one could output two point clouds instead of one: the first containing the last returns, the second containing the strongest returns. One could also add a mode flag to the point cloud indicating the kind of return.

### Your contribution
If you would like to add a feature that allows us to differentiate between the modes of return for each point, or if you have ideas how to make the code cleaner, I would be happy.

jack-oquin:
I appreciate this work, @fudger, but am not inclined to merge it any time soon. I prefer to get out a more minimal release for Jade and Indigo, first.

And, I'll be on vacation for the next two weeks, anyway.

fudger:
Thank you for letting me know, @jack-oquin. I realize that given the current architecture, this feature adds quite a few lines of code. 

If at some point, you plan to refactor the driver so that dual returns, spherical coordinates, point-wise coordinate transformations, etc. can be integrated smoothly, please let me know. I would be glad to contribute.

jack-oquin:
That's right, and I have no objection to doing that in Kinetic, but I want to make a more-minimal release to Jade and Indigo, first. Normally, I would not release to those distros at all, but Indigo needs an update so its binaries will support the VLP-16.

jack-oquin:
@fudger: your contributions are very welcome. We can discuss further, after the J and I releases are done. 

(I'll probably release that same code to Kinetic, just to get basic function. Then, we can discuss code cleanup, it's probably overdue.)

fudger:
Perfect, I am looking forward to it.

spuetz:
Hi, I'm also interested in contributing to a new optimal architecture by considering all requirements (for all laser scanners). I think that's not really difficult. The one big thing is to test the new arichtecture and to accept it. So we need all scanners and some test scenarios. We have the VLP-16 here. So what is next?

fudger:
In my opinion, the next steps consist in devising an improved architecture, then refactoring the code, and finally testing it using all scanners.

So how do we discuss the architecture? I suggest we open a new issue and compare our different approaches. I already have a first draft at hand. Unfortunately, I will not be able to participate for three weeks starting from tomorrow.

PR #76, opened 2016-02-12
Sign inversion in some equations

pomerlef:
I had to reverse engineered your driver for some calibration and, when expressing your equations with rotation matrix, realized that some signs were inverted. I could send you the equations but it's also consistent with the following:
- [Velodyne manual](http://www.velodynelidar.com/lidar/products/manual/HDL-64E%20S3%20manual.pdf) p. 44

 ```
xyDistance = distance * cosVertAngle;// — vOffsetCorr * sinVertAngle;
```

- this [calibration paper](http://cdn.intechweb.org/pdfs/23811.pdf), eq. (4)

### DIFF
--- a/velodyne_pointcloud/src/lib/rawdata.cc
+++ b/velodyne_pointcloud/src/lib/rawdata.cc
@@ -183,7 +183,7 @@ namespace velodyne_rawdata
            * was added to the expression due to the mathemathical
            * model we used.
            */
-          float xy_distance = distance * cos_vert_angle + vert_offset * sin_vert_angle;
+          float xy_distance = distance * cos_vert_angle - vert_offset * sin_vert_angle;
   
           // Calculate temporal X, use absolute value.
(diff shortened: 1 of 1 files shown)


jack-oquin:
Thanks for the contribution.

Your interpretation looks right to me. That calculation had been added last year by "Kun Li ", who is not currently reachable via github messages.

Why do you suppose that ``// - vOffsetCorr`` expression is shown in the doc as a comment, and not as part of the code?

pomerlef:
Not clear why they went into the trouble of putting it in a comment. This part ` - vOffsetCorr * sinVertAngle;` is only present in their latest documentation version. Maybe people pointed out that it was an approximation without it and they are slowly including it.

Those kind of missing elements start to make a difference with large distances. What is sure is that the parts added by "Kun Li" are mathematically correct. There was just a typo I guess.

pomerlef:
Is there anything I can do for the rest?

jack-oquin:
Not that I can think of, I've been busy on other projects and had no time for Velodyne work lately.

PR #78, opened 2016-02-22
Dynamic reconfiguration of transform_node

fudger:
### Before
Until now, transform_node does neither read parameters other than frame_id from the command line nor does it expose these parameters via dynamic reconfigure. As parameters like max_range and view_width are initialized to zero, transform_node always returns an empty point cloud:
`rosrun velodyne_pointcloud transform_node _calibration:=ros/calibration/velodyne_64e.yaml`
screenshot from 2016-02-22 10 39 24

### After
The commits in this pull request alter transform_node's behavior: Now, the node launches an reconfigure server just as cloud_node. The default configuration values are the same as in cloud_node. In addition to cloud_node, transform node exposes the frame_id parameter to reconfiguration. Now, the output looks like this:
`rosrun velodyne_pointcloud transform_node _calibration:=ros/calibration/velodyne_64e.yaml`
screenshot from 2016-02-22 10 41 40

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@
 *.tar.gz
 *~
 *.user
+doc
 TAGS
--- a/velodyne_pointcloud/CMakeLists.txt
+++ b/velodyne_pointcloud/CMakeLists.txt
@@ -10,7 +10,9 @@ set(${PROJECT_NAME}_CATKIN_DEPS
     sensor_msgs
     tf
     velodyne_driver
-    velodyne_msgs)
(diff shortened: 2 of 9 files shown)


jack-oquin:
Looks good to me. Thanks for doing this.

jack-oquin:
@fudger: many thanks for finding and fixing this significant bug. 

I made a few small changes, and some test case improvements to go with it.

Please re-run your test with the latest ``master`` if you get the chance.

PR #80, opened 2016-03-09
allow floats in min/max_intensity and make horiz_offset_correction optional

wjwwood:
I had need to import a file which contained floats instead of ints for the `min_intensity` and `max_intensity` fields in a calibration file. So the first thing this pr does is change how those values are extracted from YAML. First it tries to get them as ints, but if that throws an InvalidScalar exception, then I try to put it in a float and if that works I cast the floor the float to the int storage and warn once.

I believe this will also address https://github.com/ros-drivers/velodyne/issues/45. It will attempt to extract `horiz_offset_correction` from the YAML file and if it is not there, then it defaults to `0`, which I believe is the consensus that was reached in #45.

### DIFF
--- a/velodyne_pointcloud/src/lib/calibration.cc
+++ b/velodyne_pointcloud/src/lib/calibration.cc
@@ -73,11 +73,17 @@ namespace velodyne_pointcloud
     node[DIST_CORRECTION_X] >> correction.second.dist_correction_x;
     node[DIST_CORRECTION_Y] >> correction.second.dist_correction_y;
     node[VERT_OFFSET_CORRECTION] >> correction.second.vert_offset_correction;
-    node[HORIZ_OFFSET_CORRECTION] >> correction.second.horiz_offset_correction;
 #ifdef HAVE_NEW_YAMLCPP
+    if (node[HORIZ_OFFSET_CORRECTION])
+      node[HORIZ_OFFSET_CORRECTION] >>
+        correction.second.horiz_offset_correction;
 #else
+    if (const YAML::Node *pName = node.FindValue(HORIZ_OFFSET_CORRECTION))
+      *pName >> correction.second.horiz_offset_correction;
 #endif
     else
+      correction.second.horiz_offset_correction = 0;
+
     const YAML::Node * max_intensity_node;
 #ifdef HAVE_NEW_YAMLCPP
     max_intensity_node = node[MAX_INTENSITY];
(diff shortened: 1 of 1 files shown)


jack-oquin:
Thanks for the fixes, @wjwwood. Much appreciated!

jack-oquin:
Should we delete the ``fix_yaml_import`` branch now?

wjwwood:
My pleasure. Yeah that's fine, I deleted the branch.

velodyne_pointcloud/src/lib/calibration.cc
```diff
@@ -73,25 +73,56 @@ namespace velodyne_pointcloud
...
-      *pName >> correction.second.max_intensity;
+    if (const YAML::Node *pName = node.FindValue(HORIZ_OFFSET_CORRECTION))
+      *pName >> correction.second.horiz_offset_correction;
 #endif
     else
+      correction.second.horiz_offset_correction = 0;
+
+    const YAML::Node * max_intensity_node;
+#ifdef HAVE_NEW_YAMLCPP
+    max_intensity_node = node[MAX_INTENSITY];
```
jack-oquin: I'm getting a compile error here:
```
cannot convert const YAML::Node to const YAML::Node* in assignment
```
What's it supposed to be doing?

velodyne_pointcloud/src/lib/calibration.cc
```diff
@@ -73,25 +73,56 @@ namespace velodyne_pointcloud
...
+    }
+    else {
       correction.second.max_intensity = 255;
+    }
+
+    const YAML::Node * min_intensity_node;
 #ifdef HAVE_NEW_YAMLCPP
-    if (node[MIN_INTENSITY])
-      node[MIN_INTENSITY] >> correction.second.min_intensity;
+    min_intensity_node = node[MIN_INTENSITY];
```
jack-oquin: Same compile error here.

jack-oquin:
Which yaml-cpp did you test with?

wjwwood:
That's a good question, let me check (I certainly didn't check with both). I'll test it with both and open a fixup pull request.

jack-oquin:
Looks like Trusty has 0.5. That's mainly what matters. 

The older one should probably be deprecated.

jack-oquin:
Do we still build Indigo on Saucy? I wonder what version that had. It's so old, it's hard to find out.

wjwwood:
I think the problem was that I was using 0.3 on OS X (where we have yaml-cpp-0.3 and yaml-cpp which is 0.5). I'm testing a fix now on my Trusty VM.

jack-oquin:
Thanks. In that case, we definitely still need 0.3 support for OS X.

PR #88, opened 2016-04-20
Add dynamic latency configuration

Tones29:
This pull request enables both the transform_node and the cloud_node to manipulate the capture time of sensor data via dynamic_reconfigure. This allows the user to compensate constant time offsets. The feature is highly valuable for the synchronization of multi-sensor setups. The chosen latency time (parameter: '~latency') is subtracted from the time stamps, so that this parameter should typically be positive. For my setup, `~latency` is in the range of 20ms.

### DIFF
--- a/velodyne_pointcloud/cfg/CloudNode.cfg
+++ b/velodyne_pointcloud/cfg/CloudNode.cfg
@@ -12,5 +12,7 @@ gen.add("view_direction", double_t, 0, "angle defining the center of view",
         0.0, -pi, pi)
 gen.add("view_width", double_t, 0, "angle defining the view width",
         2*pi, 0.0, 2*pi)
+gen.add("latency", double_t,  0, "constant latency of the sensor in seconds",
+        0.0, -1.0, 1.0)
 
 exit(gen.generate(PACKAGE, "cloud_node", "CloudNode"))
--- a/velodyne_pointcloud/cfg/TransformNode.cfg
+++ b/velodyne_pointcloud/cfg/TransformNode.cfg
@@ -35,5 +35,12 @@ gen.add("frame_id",
   0, 
   "new frame of reference for point clouds", 
   "odom")
+  
+gen.add("latency",
+  pgc.double_t,
+  0,
+  "constant latency of the sensor in seconds",
+  0.0, -1.0, 1.0)
+  
 
 exit(gen.generate(PACKAGE, "transform_node", "TransformNode"))
--- a/velodyne_pointcloud/src/conversions/convert.cc
+++ b/velodyne_pointcloud/src/conversions/convert.cc
@@ -35,7 +35,7 @@ namespace velodyne_pointcloud
     dynamic_reconfigure::Server<velodyne_pointcloud::CloudNodeConfig>::
       CallbackType f;
     f = boost::bind (&Convert::callback, this, _1, _2);
-    srv_->setCallback (f);
+    srv_->setCallback (f); // Set callback function und call initially
 
     // subscribe to VelodyneScan packets
     velodyne_scan_ =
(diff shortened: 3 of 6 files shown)


jack-oquin:
Thanks for the contribution! This looks like a very useful enhancement.

jack-oquin:
Why did you add this to the point cloud conversions instead the driver?

Tones29:
Jack, you are welcome. From my understanding, the velodyne_driver receives the data from the velodyne and publishes it 'as is'. All the transformations (in space or time) are then done in the velodyne_pointcloud nodes. Furthermore, veldoyne_driver does not depend on dynamic_reconfigure so far, so I would have had to introduce an additional dependency. But I agree with you that there might also be reasons to correct the timestamp in the driver. Would you prefer this way?

jack-oquin:
I think I prefer doing it in the driver, but could perhaps be persuaded otherwise. 

My thought is that the driver sets the times, knowing that there is some positive latency, but currently without a good estimate of the delay. Note that it estimates the time for each packet, as well as the entire message. Doing it there puts all that logic in one place.

Adding dynamic reconfig to ``velodyne_driver`` isn't that hard, and might be useful for some of the other parameters, like ``npackets``. I would leave the ``model`` as a static parameter, though. No point in changing that dynamically.

What do you think? Would that be a better way?

Tones29:
That's fine for me as well, I think both ways have their advantages and disadvantages.  I will try to move the changes to the driver next week. What code style are you using for this repository and do you have by any chance the eclipse imports for this codestyle?

Tones29:
Jack, I managed to adjust the velodyne_driver today. Please review pull request #89. Note, that I renamed the parameter from `latency` to `time_offset`, as this naming scheme is used in several other packages. In contrast to `latency`, `time_offset` is added to the time stamp, so this value should typically be negative. This is more consistent with the implementation in other packages such as urg_node (hokuyo) and the openni drivers.

jack-oquin:
Code style: too many hands in this code already. :smile: Just try to copy the style of the files you update.

I'd love to enforce some kind of roslint via a CI test, but there are too many active updaters who would be disrupted by re-formatting the source to fit a standard.

Re: ``time_offset`` thanks for checking consistency with other packages. :+1:

chukcha2:
It seems to be a very nice feature but it requires one to know what the latency (between hw packet capture and ROS driver packet capture) is and how it the latency changes throughout run time. For Velodyne, how are you able to find out what the latency is?

Tones29:
chukcha2, based on my observations, the latency of the velodyne does not vary significantly during runtime. The feature allows to consider latency times, but does not offer any calibration methods.

I personally use the latency adjustment in order to calibrate the velodyne against an imu that is mounted on top of the velodyne. Therefore, I mount velodyne + imu on a tripod in the middle of a room and visualize the velodyne pointcloud in the imu odom frame. Adjust the settings of your viewer (e.g. rviz) to top-down view and observe the position of the corners of the room when you turn the sensor in direction of the z-axis. If the latency correction is correct, the corners should stay stable. Please remark, that this method does not estimate the absolute latency of the velodyne but the latency of the velodyne in respect to the imu data.

HiOtto:
> This pull request enables both the transform_node and the cloud_node to manipulate the capture time of sensor data via dynamic_reconfigure. This allows the user to compensate constant time offsets. The feature is highly valuable for the synchronization of multi-sensor setups. The chosen latency time (parameter: '~latency') is subtracted from the time stamps, so that this parameter should typically be positive. For my setup, `~latency` is in the range of 20ms.

I am doing a synchronization between velodyne and IMU, but my estimation results in 20ms delay between velodyne and IMU. So I want to consult how did you get your default value (20ms range)? Is that a true value of velodyne delay after synchronization?

PR #96, opened 2016-07-01
updated VLP-16 packet rate from user manual.

priyankadey:
Also verified with sensor. It reduced overlap in the pointcloud
User manual:
http://velodynelidar.com/docs/manuals/63-9243%20Rev%20B%20User%20Manual%20and%20Programming%20Guide,VLP-16.pdf
Page #5

### DIFF
--- a/velodyne_driver/src/driver/driver.cc
+++ b/velodyne_driver/src/driver/driver.cc
@@ -55,7 +55,7 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
     }
   else if (config_.model == "VLP16")
     {
-      packet_rate = 781.25;             // 300000 / 384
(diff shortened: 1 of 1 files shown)


jack-oquin:
Thanks for the fix!

chukcha2:
I guess this means that the sampling rate is 289,536 samples/s and not 300,000 samples/s.

The manual also says that there are

1: 12 data blocks / packet  (page 21)
2: 32 points / data block (page 21)

Upon closer look the manual does not specify sampling rate. The spec sheet (http://velodynelidar.com/docs/datasheet/63-9229_Rev-C_VLP16_Datasheet_Web.pdf) says "~300,000".

Given that there are 754 packets/s, the real sample rate must be 289,536 samples/s

PR #101, opened 2016-08-09
velodyne_driver/src/lib/input.cc : fix for device_ip filter

teosnare:
Fix for device_ip filter in InputSocket: initialization of devip_ for correct ip filtering in InputSocket::getPacket.


jack-oquin:
Thanks for the fix! (xref #107)

PR #103, opened 2016-08-11
Fix misleading typecasts.

fudger:
intensity and VPoint::intensity are both of type float.


jack-oquin:
Thanks for the contribution, @fudger.

Issue #90, opened 2016-04-29
add new parameter arguments to various launch files


jack-oquin:
At least these parameters are missing from some of the launch scripts:

 * port
 * device_ip
 * rpm

There may be others.

Affected scripts include:

 * ``velodyne_driver/launch/nodelet_manager.launch``
 * ``velodyne_pointcloud/launch/VLP16_points.launch``
 * ``velodyne_pointcloud/launch/32e_points.launch``

There are probably others.

See: http://answers.ros.org/question/232949/different-rpmportreturn-type-velodyne-vlp16/?answer=233290#post-id-233290

ZepherusFF:
Hi Jack, Thanks for the answer/opening this up. 

I think I'm gonna need more details x) 

Should I setup my launch file like this : 
<launch>

  
  <arg name="pcap" default="" />
  <arg name="calibration" default="$(find velodyne_pointcloud)/params/VLP16db.yaml"/>
  <arg name="min_range" default="0.4" />
  <arg name="max_range" default="130.0" />


  
  <include file="$(find velodyne_driver)/launch/nodelet_manager.launch">
    <arg name="model" value="VLP16"/>
    <arg name="pcap" value="$(arg pcap)"/>
**   <arg name="return_type" value="Dual" />
   <arg name="rpm" value="1200" />
   <arg name="port" value="myport_number" />
   <arg name="IP" value="my_ip_number" />
**
  </include>

  
  <include file="$(find velodyne_pointcloud)/launch/cloud_nodelet.launch">
    <arg name="calibration" value="$(arg calibration)"/>
    <arg name="min_range" value="$(arg min_range)"/>
    <arg name="max_range" value="$(arg max_range)"/>
  </include>

</launch>


and my nodelet manager like this : 

<launch>

  
  <node pkg="nodelet" type="nodelet" name="velodyne_nodelet_manager"
        args="manager" />
  <arg name="model" default="64E" />
  <arg name="pcap" default="" />
  <arg name="read_once" default="false" />
  <arg name="read_fast" default="false" />
  <arg name="repeat_delay" default="0.0" />
**  <arg name="return_type" default="Dual" />
  <arg name="rpm" default="1200" />
  <arg name="port" default="2378" />
  <arg name="IP" default="192.168.1.201"/>**
  <arg name="frame_id" default="velodyne" />
  <node pkg="nodelet" type="nodelet" name="driver_nodelet"
[...]

PR #104, opened 2016-08-11
Add more options in launch files.

altrouge:
Relative to #90 

Add the following options to VLP16 and hdl32 launch files:
- rpm, device_ip, port, read_once, read_fast, repeat_delay

In the 32e_points, there was also no min_range and max_range, I put the same as the vlp16 (though I'm not certain it's correct, 0.4 to 130), and pcap.

Finally, I added the port and device_ip to the nodelet manager.

I am not very familiar with ROS, so am not sure overriding defaults parameters with the same defaults is the best way for some of them, but couldn't find any other way. Please correct me if I am mistaken.

### DIFF
--- a/velodyne_driver/launch/nodelet_manager.launch
+++ b/velodyne_driver/launch/nodelet_manager.launch
@@ -9,27 +9,27 @@
   <!-- start nodelet manager and load driver nodelet -->
   <node pkg="nodelet" type="nodelet" name="velodyne_nodelet_manager"
         args="manager" />
+  <arg name="device_ip" default="" />
+  <arg name="frame_id" default="velodyne" />
   <arg name="model" default="64E" />
   <arg name="pcap" default="" />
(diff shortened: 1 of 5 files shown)


jack-oquin:
That looks good to me. It is annoying having to duplicate the parameter defaults in the scripts, but it's necessary because they are setting the parameters explicitly, overriding their built-in defaults.

My only suggestion is that now the scripts have so many args and params, it would help to alphabetize them so we can quickly see what's not there.

The CI failure looks like an internal Travis problem. I don't think it's related to these changes in any way.

Thanks for contributing! Much appreciated.

altrouge:
I tried to arrange the parameters by type (pcap, cloud nodelet, ...), but it is indeed more readable when arranged alphabetically.

jack-oquin:
Thanks for doing this, @altrouge! It will help a lot of people.

PR #105, opened 2016-08-16
Remove unused constants.

fudger:
DISTANCE_MAX and DISTANCE_MAX_UNITS are not used anywhere in the code.

Furthermore, using them would lead to errors as both VLP-64 manuals state that returns above 120 m should not be used. The VLP-32 manual allows 70 m as the maximum valid sensor range.


jack-oquin:
Thanks for the fixes, @fudger!

PR #110, opened 2016-12-08
Added velodyne_laserscan package

kmhallen:
Extract a single ring from the [PointCloud2](http://docs.ros.org/api/sensor_msgs/html/msg/PointCloud2.html) topic ```velodyne_points``` and publish a [LaserScan](http://docs.ros.org/api/sensor_msgs/html/msg/LaserScan.html) topic to ```scan```

* Fast and efficient algorithm using ```sensor_msgs/point_cloud2_iterator.h``` (no PCL)
* Automatically select ring for known devices (VLP-16, HDL-32E, HDL-64E)
* Lazy subscribers that don't even subscribe until the scan output is requested
* Integrated into existing launch files

### DIFF
--- a/velodyne/package.xml
+++ b/velodyne/package.xml
@@ -18,6 +18,7 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <run_depend>velodyne_driver</run_depend>
+  <run_depend>velodyne_laserscan</run_depend>
   <run_depend>velodyne_msgs</run_depend>
   <run_depend>velodyne_pointcloud</run_depend>
 
--- a/velodyne_laserscan/CMakeLists.txt
+++ b/velodyne_laserscan/CMakeLists.txt
@@ -1,0 +1,46 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(velodyne_laserscan)
+
+find_package(catkin REQUIRED COMPONENTS
+  roscpp
+  nodelet
+  sensor_msgs
+  dynamic_reconfigure
+)
+
+generate_dynamic_reconfigure_options(
+  cfg/VelodyneLaserScan.cfg
+)
+
+catkin_package()
+
+include_directories(
+  ${catkin_INCLUDE_DIRS}
+)
+
+add_library(${PROJECT_NAME}
+  src/VelodyneLaserScan.cpp
+  src/nodelet.cpp
+)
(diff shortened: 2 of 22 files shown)


jack-oquin:
This looks useful, and is not anything I had ever considered.

Some questions:
* Why is the new nodelet not in the new package?
* Why did you modify some of the existing velodyne_pointcloud launch files?
* What ring do you automatically select?
* Are you willing to write documentation for the new features?
* Are you willing to write unit tests for the new features?

kmhallen:
Answering Jack's questions:

* Why is the new nodelet not in the new package?
    * It could probably be moved. I didn't want to create a dependency loop where both the velodyne_pointlcoud and velodyne_laserscan packages depend on each other.
* Why did you modify some of the existing velodyne_pointcloud launch files?
    * Because the laserscan node does nothing (not even subscribe) until the output is requested, I decided to make it part of the system by default.
* What ring do you automatically select?
    * The ring that is closest to zero angle. Otherwise, a LaserScan doesn't make sense. The LaserScan message does not account for a vertical angle.
* Are you willing to write documentation for the new features?
    * Yes, I can document this on the wiki.
* Are you willing to write unit tests for the new features?
    * Yes, I will have someone working for me write unit tests :)

This is comparable to the but_velodyne_proc laserscan node. The but_velodyne implementation looks at z height instead of ring number, and is less efficient (uses PCL).  
http://wiki.ros.org/but_velodyne_proc#Laserscan_Node

I have seen several users want to use their velodyne sensor with 2D software like costmap_2d and gmapping.  
http://answers.ros.org/question/249537/  
http://answers.ros.org/question/239856/

volkandre:
This feature is great, but I have some trouble getting a 360 degree laserscan out of my pointcloud. As you can see in the image, I only get a 180 degree laserscan of the right half of one ring of my pointcloud. What am I doing wrong?

velo_laser

JWhitleyAStuff:
@kmhallen - Would you mind pulling in the last couple of changes from the main repo to yours? I'll then validate this on a couple of sensors and merge it in provided everything looks good.

kmhallen:
The latest commits should be merged now

JWhitleyAStuff:
Tested on a 32 and a 16. Negligible overhead was generated for the 32 so I don't see this being a big issue. Looks good to me. Merging.

PR #111, opened 2016-12-20
Added an interface to set up raw data processing offline

tstoyanov:
At our lab we have had this local change for a while now and thought it might be useful to push out. It's a small tweak to allow us to setup the driver for offline processing from a locally defined calibration file. This method is useful when processing data offline from a bag file, without starting any ros master.


jack-oquin:
Looks useful, thanks for contributing!

PR #126, opened 2017-09-28
Updated "cut at specified angle feature" with latest master version.

volkandre:
I noticed, that the "cut at specified angle feature" branch was no longer compatible with the latest version of the master branch, so I updated it.


JWhitleyAStuff:
@volkandre - Thanks for updating this branch! A couple of PRs have been pulled in since you did this. Would you mind merging the latest with yours? Also, some questions regarding this feature:

What exactly is meant by "cut_angle"?
If this is intended to "cut" the output at a specified angle, I'm guessing it's from 0 degrees to this angle in degrees? If so, wouldn't it make more sense to have a "cut_start_angle" and "cut_end_angle" (the HDL-64E allows you to set this internally but the other sensors do not).

Thanks,

Josh

velodyne_driver/src/driver/driver.cc
```diff
@@ -172,7 +185,7 @@ bool VelodyneDriver::poll(void)
 
   // publish message using time of last packet read
   ROS_DEBUG("Publishing a full Velodyne scan.");
-  scan->header.stamp = ros::Time(scan->packets[config_.npackets - 1].stamp);
+  scan->header.stamp = scan->packets[config_.npackets - 1].stamp;
```
cfneuhaus: feels like this should be `scan->packets.back().stamp;`. A scan with cut_angle set can have less than `npackets` packets.

cfneuhaus:
@JWhitleyAStuff The velodyne assembles multiple packets into one scan. The 'cut angle' specifies an angle at which a revolution of the velodyne is considered 'complete'. You generally want this to be behind the vehicle for example, because you might otherwise have 'jumps' in the area in front of you.

velodyne_driver/src/driver/driver.cc
volkandre: You are exactly right. Thanks for finding this bug. The bug caused a duplication of timestamps. I will fix this immediately.

volkandre:
@JWhitleyAStuff I updated my pull request, so it is now compatible with the latest version of the master branch. I've successfully tested it on a VLP-16 LIDAR. 
Thanks to @cfneuhaus for the explanation. Please also see https://github.com/ros-drivers/velodyne/issues/33 for screenshots and a detailed description of the underlying problem with data from velodyne LIDARs.
It would be great, if this feature could be pulled into the master branch, because it is tedious to keep a feature branch in sync with the master branch. And the feature works perfectly optional. It is only activated, if someone defines a "cut_angle" >= 0.

JWhitleyAStuff:
@volkandre - Sorry for the delay. Lots of crazy stuff going on around the holidays. So, the branch has been updated on the primary repository. I only have one more piece of feedback - Based on REP 103 (http://www.ros.org/reps/rep-0103.html), would you mind making this an angle in radians instead of degrees? I get the need for this feature but I want to make sure we're compliant with the standard REPs. I've looked over the code changes and they look fine so if you add this one thing, I'll merge it into master. Thanks.

PR #127, opened 2017-09-28
Add VLP16 Puck Hi-Res support

kriskozak:
Added a calibration file for the Hi-Res version of the VLP16, which has a +/-10 degree vertical field of view rather than the +/-15 degree field of view for the standard VLP16 model.

For reference:
* http://velodynelidar.com/vlp-16-hi-res.html
* http://velodynelidar.com/vlp-16.html


pjreed:
:+1: 
We could use this for some of our projects, so it would be great to get this merged in and get a new release.

PR #130, opened 2017-11-10
added a requirement

huckl3b3rry87:


JWhitleyAStuff:
Instead of providing Debian/Ubuntu-specific instructions on how to fix the issue, would you mind just creating a request that adds libpcap as a build and runtime requirement so it is compatible with rosdep and gives a more descriptive error on build? I planned to do this soon but I wanted to get a release out first.

PR #136, opened 2017-12-12
Use std::abs instead of abs

stsundermann:
abs is the c version which returns an integer. This is probably not intended here, so use the templated std::abs function.


JWhitleyAStuff:
Thanks for the fix!

PR #149, opened 2018-02-15
Code and files are added for VLP32C from Velodyne

magiccjae:
I am doing this by the request of Mike Parkin.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@ Overview
 ========
 
 Velodyne_ is a collection of ROS_ packages supporting `Velodyne high
-definition 3D LIDARs`_ .
+definition 3D LIDARs`_.
 
 **Warning**::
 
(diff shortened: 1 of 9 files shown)


JWhitleyAStuff:
Sorry for not getting to this for a few days. So, I have a couple of issues with this set of changes:

1. It does not appear to include the "cut_at_angle" feature which was added recently. Would you mind pulling from master to help keep the merge clean?
2. This commit is much more of an "overhaul" than just adding support for the VLP-32C - especially in rawdata.h/.cc. Since we have stable versions out for every ROS release except the latest, I am OK talking about a general driver overhaul but I'm really not OK including it under the guise of adding support for a new sensor. If the changes you've made are necessary to support the 32C, would you mind making a PR with just those "core" changes first? The community can then discuss those changes and then add the additional support for the 32C in a separate PR.

We are all certainly happy to see the Velodyne team take an interest in supporting ROS with contributions to this driver but we want to make sure everything is transparent and that it is still easy for the community to maintain it. Thanks.

JWhitleyAStuff:
@willdzeng - I have been waiting for a response from @magiccjae or Velodyne in general. No word yet. It is on our list internally to contact them directly but we haven't gotten to it yet. I will up the priority. Thanks for your patience.

JWhitleyAStuff:
Since there has been no movement on this PR in 2 months and Velodyne doesn't seem to be responding, I'm closing this PR.

PR #151, opened 2018-02-23
Raise diagnostic when no packets are received

Axel13fr:
Added a periodic update of the diagnostics so that when no data is received at all from the Velodyne, a diagnostic information will be published. The previous implementation would publish diagnostics only on packet reception.


JWhitleyAStuff:
Axel - Thanks for adding this. It is convention to provide a variable name for the arguments passed to functions. Would you mind adding this? Then I will give it a test and merge it in. Thanks.

Axel13fr:
Sure, there you go.

Issue #142, opened 2018-01-29
Error in packet rate for the VLP-32C


mpitropov:
With the motor RPM set to 600 and a packet rate of 1808, I get around 8.3 hz on the /velodyne_packets topic.

With a packet rate of 1508:
rostopic hz /velodyne_packets 
subscribed to [/velodyne_packets]
average rate: 9.987
	min: 0.100s max: 0.100s std dev: 0.00011s window: 10
average rate: 9.984
	min: 0.100s max: 0.100s std dev: 0.00009s window: 19
average rate: 9.982
	min: 0.100s max: 0.100s std dev: 0.00008s window: 29
average rate: 9.982
	min: 0.100s max: 0.100s std dev: 0.00007s window: 39

Has anyone else tested the packet rate set for this model? I think that the packet rate should be at around 1508.

jarvis-huang:
I experienced the problem as @mpitropov . Any fix would be highly appreciated.

asimay:
@mpitropov , hi , 32C has 60,0000 points per second, it is not same with 32E. so you should change the packet rate, for 32C the packet rate is not 1808.

asimay:
@jarvis-huang , hi, are you udacity's employee? I'm udacity's SDC students and mentor. :)

JWhitleyAStuff:
Fixed by #154.

PR #154, opened 2018-03-09
Fix packet rate for the Velodyne 32C

awesomebytes:
Following https://github.com/ros-drivers/velodyne/issues/142 adding the correct packet_rate


JWhitleyAStuff:
Looks good, compiles, following the discussion, I see no reason not to merge.

PR #155, opened 2018-03-09
Avoid unnecessary computation in RawData::unpack

facontidavide:
Hi, I noticed that the method __RawData::unpack_vlp16__ could be optimized.

1) __calibration_.laser_corrections__ is a std::map. This means that the find() operation is relatively expensive and not really necessary for each point. I stored the pointer to the laser correction in a simple array with 16 elements.

2) Since a point that won't pass the __pointInRange(distance)__ condition is not stored, I do this check as soon as possible to avoid unnecessary computation.

The speed up improvement on my machine is considerable. The CPU usage decreases from 25% to 15% (I have a VLP16).

Cheers

Davide


JWhitleyAStuff:
It looks like this could also be done for the other lasers. Would you mind pulling in the latest `master` updates and modifying RawData::unpack in a similar manner? I will test on several Velodynes at my office once this is done.

facontidavide:
Take a look to f59ea36b1e63f0b7de3345ac9d8eb10bcb8fe264

Is this what you meant?

I am not able to test it on here, so... fingers crossed!

If it works, can you please tell me how much difference in CPU usage you notice on you computer, using an app such as __top__ or __htop__ ? 

Cheers

JWhitleyAStuff:
Derp. Totally missed that one. Thanks. I'll give it a test as soon as I can get some time on the Velodynes.

facontidavide:
Looking forward for it. With so many velodyne all around the world, image how much energy can be saved :P

facontidavide:
for the records. This can be further improve if we don't use std::map in first place, instead of converting from std::map to std::vector in every call of unpack.

Therefore, if you test these changes and it works, please let me know before merging.

facontidavide:
799dd49c224589b6e55133cfefe2eceb07187199 this solve the source of the problem ;) 

Less code and same result.

facontidavide:
@JWhitleyAStuff this is a friendly ping? Did you have the opportunity to test this PR?

JWhitleyAStuff:
Sorry, we have not. We have a handful of high-priority projects so this probably won't get tested here until the end of June.

JWhitleyAStuff:
@facontidavide - Would you mind resolving the conflicts generated when I pulled in another PR? I'll also go ahead and pull this one in after that. My testing is taking far too long and we'll rely on the community for testing this one if you aren't able to.

facontidavide:
I will be happy to solve the conflict, BUT I do think that the changes should be tested on the hardware by SOMEONE before being merged.

facontidavide:
conflict fixed, but I didn't have the opportunity to test it, because my Velodyne is broken :(

JWhitleyAStuff:
@facontidavide - Please see PR #194. This makes it easier for me (and others) to test the changes. Sorry again for the lack of recent updates. I'll see if I can get this moved up in priority.

facontidavide:
ok, if you need any action from my side let me know

PR #164, opened 2018-05-10
Adding VLP-32C support.

JWhitleyAStuff:
Based on work done by @rockcdr. Adds distance_resolution calibration
value to support 0.004m distance resolution for VLP-32C.

Needs testing.


velodyne_pointcloud/include/velodyne_pointcloud/calibration.h
```diff
@@ -64,9 +65,10 @@ namespace velodyne_pointcloud {
   public:
 
     Calibration(bool info=true):
-      initialized(false), ros_info(info) {}
+      distance_resolution_m(0.002f), initialized(false), ros_info(info) {}
     Calibration(const std::string& calibration_file,
                 bool info=true):
+      distance_resolution_m(0.002f),
```
mikepurvis: Is this supposed to be overridable using rosparams at runtime? Where is the logic for that?

JWhitleyAStuff: It's a value that's included in the calibration YAML files. See the changes to velodyne_pointcloud/scripts/gen_calibration.py for the additional parsing logic.

JWhitleyAStuff:
For anyone watching this issue, I'm sorry it has been delayed so long. I am getting clarification from Velodyne on the distance resolution. It appears that it may not be the same across the whole distance field of the sensor (better resolution at <50m and worse at >50m).

willdzeng:
Thanks for doing this,
I wonder if the packet rate needs to be changed for vlp32c something like this in the other PR?
https://github.com/ros-drivers/velodyne/pull/149/files#diff-5707fd8c8886aa445fb461efc9d17a2fR59

JWhitleyAStuff:
@willdzeng - Looking at the branch for this PR (`maint/vlp_32c_support`), this change was incorporated. I see the following at line 58 of `velodyne_driver/src/driver/driver.cc`:

    packet_Rate = 1507.0;

willdzeng:
I see, Didn't realize that it's already checked in.

PR #177, opened 2018-07-30
Feature/wrap pointcloud data

C-NR:
I'm proposing the use of a small abstract wrapper class for the pointcloud to be filled, so that arbitrary data structures such as range images, octo trees, of cause pointclouds , ...  can be directly build using the existing code for laser beam correction. Any data structure just inherits from the abstract class and implements the method "addPoint(const float& x, const float& y, const float& z, const uint16_t& ring, const uint16_t& azimuth, const float& distance, const float& intensity)" in which the pointcloud or other data structure adds a new point to itself.


JWhitleyAStuff:
Has this code been tested on a sensor? If so, which one? And was it tested with strongest, last, and dual-return modes?

C-NR:
Hi Joshua,

the code has been tested with a Velodyne HDL-64 S2.1 in last return mode 
and on a VLP-16 on all three modes so far.

Best,
Claas

Am 2018-08-02 05:47, schrieb Joshua Whitley:

Issue #174, opened 2018-06-20
Cut at specified angle leads to deadlock


sts-thm:
The ```cut_angle``` parameter and functionality introduced by #138 leads to a deadlock inside a ```while(true)```-loop, for specific values of the ```cut_angle``` parameter in relation to the FOV specified in the velodyne's user interface.
I found out about this issue using a VLP16 and commit ef2345eb0633360c2ab80c231bb65f45866df82e of this repository.
There are several possible configurations for the relation between the two angles (Start, Stop) specifying the FOV of the velodyne in the velodyne's user interface and the ```cut_angle``` parameter (c):

1. 0° < Start < c < Stop < 360°
2. 0° < c < Start < Stop < 360°
3. 0° < Start < Stop < c < 360° <- bad case 
4. 0° < Stop < c < Start < 360°
5. 0° < c < Stop < Start < 360°
6. 0° < Stop < Start < c < 360°

For the 3rd configuration the ```break``` in the ```while(true)``` (https://github.com/ros-drivers/velodyne/blob/master/velodyne_driver/src/driver/driver.cc#L189) is never reached and the loop will never stop.
My solution to this was to replace the lines 180-186 in https://github.com/ros-drivers/velodyne/blob/master/velodyne_driver/src/driver/driver.cc with:
``` cpp
      //if first packet in scan, there is no "valid" last_azimuth
      if (last_azimuth == -1) {
  	 last_azimuth = azimuth;
      	 continue;
      }
      if((last_azimuth < config_.cut_angle && config_.cut_angle <= azimuth)
      	 || ( config_.cut_angle <= azimuth && azimuth < last_azimuth)
      	 || (azimuth < last_azimuth && last_azimuth < config_.cut_angle))
```
I tested this with the same configurations and it worked.
If you want me to, I can submit a pull request for my fix.

JWhitleyAStuff:
Sorry I didn't get to this sooner. I never saw the notification. Please do submit a PR and I will get it tested/reviewed.

PR #178, opened 2018-08-01
Bugfix issue #174

sts-thm:
This pull request should be able to fix issue #174.

### DIFF
--- a/velodyne_pointcloud/src/lib/rawdata.cc
+++ b/velodyne_pointcloud/src/lib/rawdata.cc
@@ -147,7 +147,7 @@ namespace velodyne_rawdata
                        VPointCloud &pc)
   {
     ROS_DEBUG_STREAM("Received packet, time: " << pkt.stamp);
-    
+
     /** special parsing for the VLP16 **/
     if (calibration_.num_lasers == 16)
     {
@@ -317,6 +317,7 @@ namespace velodyne_rawdata
   {
     float azimuth;
     float azimuth_diff;
+    int raw_azimuth_diff;
     float last_azimuth_diff=0;
     float azimuth_corrected_f;
     int azimuth_corrected;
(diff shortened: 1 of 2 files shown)


JWhitleyAStuff:
Has this code been tested on a sensor? If so, which one? And was it tested with strongest, last, and dual-return modes?

sts-thm:
The code has been tested on a VLP-16 on strongest return mode.

velodyne_pointcloud/src/lib/rawdata.cc
```diff
@@ -340,7 +341,22 @@ namespace velodyne_rawdata
       // Calculate difference between current and next block's azimuth angle.
       azimuth = (float)(raw->blocks[block].rotation);
       if (block < (BLOCKS_PER_PACKET-1)){
-        azimuth_diff = (float)((36000 + raw->blocks[block+1].rotation - raw->blocks[block].rotation)%36000);
+	raw_azimuth_diff = raw->blocks[block+1].rotation - raw->blocks[block].rotation;
+        azimuth_diff = (float)((36000 + raw_azimuth_diff)%36000);
+	// some packets contain an angle overflow where azimuth_diff < 0 
+	if(raw_azimuth_diff < 0)//raw->blocks[block+1].rotation - raw->blocks[block].rotation < 0)
```
wenbin1989: What's the purpose of this part? I don't think there will be an angle overflow here.

@sts-thm @JWhitleyAStuff

sts-thm: This pull request fixed the issue linked above. Please refer to https://github.com/ros-drivers/velodyne/issues/174#issue-334135152. If you do still have any questions feel free to ask away.

wenbin1989: Thanks for reply. I understand the change of driver.cc, which you mentioned in the issue. But I couldn’t understand the change of rawdata.cc. It will be nice if you can explain a little, thanks.

sts-thm: The packages of the velodyne in certain modes specify the azimuth only for every second group of points (16 or 32, depending on the velodyne). The azimuth update for the group of points in between is calculated as half of the difference between the previous group's azimuth and the next group's azimuth. However if the azimuth 0° is inside your package the difference will be the reflex angle (> 180°) and not its complement, which we would need.

wenbin1989: In the case azimuth 0° is inside the packet, let's say the previous group's azimuth is 359°, the next group's azimuth is 1°.

`azimuth_diff = (float)((36000 + raw->blocks[block+1].rotation - raw->blocks[block].rotation)%36000);` is 2°, which I think is what we need?

PR #186, opened 2018-09-20
Portable handling of ECANCELED

moooeeeep:
On my machine (Ubuntu) the error code 995 does not mean a thing.

I hope I translated it correctly to the corresponding Boost.Asio code...


JWhitleyAStuff:
This is the correct implementation. According to boost/asio/error.hpp, on POSIX systems, the error ECANCELED is used (error 125: Operation Cancelled) and on Windows systems, the error ERROR_OPERATION_ABORTED is used (error 995: for IO cancellation). Thanks for the PR!

PR #189, opened 2018-09-22
Fix malformed plugin description XML

kveretennicov:
ROS pluginlib only recognizes multiple <library> elements if they are under
<class_libraries> XML root. It silently ignores malformed XMLs with multiple
<library> "root"s and just reads the first one, due to relaxed way tinyxml2 does
parsing. Though if you do `rosrun nodelet declared_nodelets`, the issue is
reported properly.

See also similar issue in https://github.com/ros-perception/perception_pcl/issues/131


JWhitleyAStuff:
Thanks for the pull request! I see that your syntax is correct based on http://wiki.ros.org/pluginlib/PluginDescriptionFile so I'm happy to pull this in.

Issue #198, opened 2018-12-05
Testing and Support for Dual-Return Mode


JWhitleyAStuff:
Dual-return mode does not seem to be properly implemented (as mentioned by #109 and shown by the problems enabling it causes in #193). This is a reminder item to eventually add full, working support for dual-return mode to the driver.

andre-nguyen:
@JWhitleyAStuff I'd like to look into this, but it is unclear to me how this would be implemented such that it works correctly for all lidar models. Any pointers?

JWhitleyAStuff:
This isn't by any means the only way but, if I was handling this, I would do the following:

- Create a launch file parameter called "dual_return_mode" with true/false values.
- Modify the `packet_rate` values for each sensor in `velodyne_driver` in driver.cc if the above launch parameter is true. For the HDL-64E S3, see page 50 of this document: https://velodynelidar.com/lidar/products/manual/HDL-64E%20S3%20manual.pdf For the HDL-64E S2, see this document: https://www.velodynelidar.com/lidar/products/manual/63-HDL64E%20S2%20Manual_Rev%20D_2011_web.pdf For the VLP-32C, see page 50 of this document: https://velodynelidar.com/docs/manuals/63-9325%20REV%20B%20MANUAL,USERS,VLP-32.pdf For the HDL-32E, see page 54 of this document: https://velodynelidar.com/docs/manuals/63-9113%20REV%20M%20MANUAL,USERS,HDL32E.pdf For the VLP-16, see page 50 of this document: https://velodynelidar.com/docs/manuals/63-9243%20REV%20D%20MANUAL,USERS,VLP-16.pdf
- The function `RawData::unpack` in rawdata.h and rawdata.cc in `velodyne_pointcloud` is where the data are unpacked from the driver-published scans. Unfortunately, some of the values here are set as constants (like the size of SCANS_PER_BLOCK which is used in calculating the BLOCK_DATA_SIZE which is used in the raw_packet_t struct) which will need to be changed to variables and doubled if the parameter is true.
- Create two point clouds if the above parameter is true.
- Either publish the usual `velodyne_points` if the above launch parameter is false or publish `velodyne_points_last` and `velodyne_points_strongest` if it is true.

Here are some documents that should help with understanding the modifications to the data structures in dual return mode:

[...]

PR #200, opened 2018-12-06
add support for 64E_S3

xiesc:
add support for 64E_S3. And the current version discards the "Last return" in S3 dual return systems. I will add the support for all mode in dual return latter when I confirm the protocol.

### DIFF
--- a/velodyne_pointcloud/src/lib/rawdata.cc
+++ b/velodyne_pointcloud/src/lib/rawdata.cc
@@ -89,7 +89,9 @@ inline float SQR(float val) { return val*val; }
 
   /** Set up for on-line operation. */
   int RawData::setup(ros::NodeHandle private_nh)
-  {
+  { 
+    // get velodyne lidar model type
+    private_nh.getParam("model", config_.model);
     // get path to angles.config file for this device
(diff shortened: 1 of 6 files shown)


velodyne_pointcloud/src/lib/rawdata.cc
```diff
@@ -175,6 +177,14 @@ inline float SQR(float val) { return val*val; }
...
       // upper bank lasers are numbered [0..31]
       // NOTE: this is a change from the old velodyne_common implementation
+      if (config_.model == "64E_S3")
+      {
+        if (i % 4 < 1) {       //only return strongest return NOTE: for 64E_S3 dual return mode.
+          continue;
+        }
+      }
+
+
```
JWhitleyAStuff: I have a problem with assuming that the 64E_S3 is in dual-return mode when all other models currently make the assumption that this *specifically* is not the case. Adding support for dual-return mode is in the pipeline (see #198) but is not currently supported by any other model.

JWhitleyAStuff:
Please remove the code for dual-return mode and add an example yaml config file and I'd be OK approving this addition.

PR #203, opened 2018-12-17
fix a bug

xiesc:
I am sorry for that I found a mistake when I double check the code. When I debugged the code last week, I changed the **line 182**. But, I forgot to change it back. Now the version is right! this line is used to ignore the redundant date when use the normal mode "Strongest". My apology.

### DIFF
--- a/velodyne_pointcloud/src/lib/rawdata.cc
+++ b/velodyne_pointcloud/src/lib/rawdata.cc
@@ -179,7 +179,7 @@ inline float SQR(float val) { return val*val; }
       // NOTE: this is a change from the old velodyne_common implementation
       if (config_.model == "64E_S3")
       {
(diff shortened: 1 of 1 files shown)


JWhitleyAStuff:
@xiesc - I didn't intend to merge that PR. I was looking at the wrong commits when I did. I commented in the code that I am not OK with making the assumption that the S3 is always running in dual-return mode. All of the other code assumes the opposite - that the sensor is running in either strongest-return or last-return mode. Please make a new PR and do one of the following:

1. Add code to handle dual-return mode for all sensors or;
2. Remove the code in your PR that handles dual-return mode.

xiesc:
I have removed the code of dual return mode. And I will add code to handle dual-return mode for all sensors latter.

xiesc:
There are some conflicts. I will create a new PR.

PR #205, opened 2018-12-18
support for 64E-S3

xiesc:
The support for 64E-S3. An example .pcap file can be found in https://github.com/xiesc/example-velodyne-64E-pcap-file


JWhitleyAStuff:
Thank you.

PR #220, opened 2019-03-18
Transform static variable into member

nbussas:
I had problems running multiple drivers when I used a nodelet manager. So I turned 'last_azimuth' into a member, which solved the issue for me. Also it's a cleaner design.


JWhitleyAStuff:
I agree and I'm OK with merging this. However, I don't see how this is related to running multiple instances of the driver.

trainman419:
The `static` variable is really a global, so it's shared among all nodelets that are launched within the same process.

PR #222, opened 2019-03-26
Use time within the packet

mpitropov:
I'm opening up a couple of PRs for changes we have made to our internal driver so that we can use the public driver again.

This is the first one and it will increase the accuracy of the ros message timestamp by using the GPS provided time. I'm thinking I could add a param to the launch files to enable GPS time instead of removed the prior functionality?


JWhitleyAStuff:
@mpitropov - I agree with your suggestion. Previous bag files used the estimated time and some user software may need to make assumptions that the time is always linear, which may not be the case using GPS-based time (e.g. the Velodyne starts out without a GPS signal and gains one, which "corrects" the time). Please add the ability to "enable" this feature with an additional parameter.

PR #223, opened 2019-03-26
Add fixed frame and use ros message time within transform node

mpitropov:
I'm opening up a couple of PRs for changes we have made to our internal driver so that we can use the public driver again.

The current implementation uses "ros::Time(0)" but we made it so that it corrects the distortion within the lidar sweep by using the message time. A fixed_frame_id was added but it might be better to remove it and rename frame_id to be fixed_frame_id. The main change is that ROS nodes that receive this message should be the ones to transform it to the right time and the driver shouldn't have to do that.


velodyne_pointcloud/cfg/TransformNode.cfg
```diff
@@ -33,7 +33,13 @@ gen.add("view_width",
...
   pgc.str_t, 
   0, 
-  "new frame of reference for point clouds", 
+  "new frame of reference for point clouds",
+  "velodyne")
+
+gen.add("fixed_frame_id",
+  pgc.str_t,
+  0,
+  "inertial reference frame for point clouds",
```
JWhitleyAStuff: Would you have a problem changing these to "frame_id" (as to not disturb current software) and "intertial_frame_id"? I think using the term "fixed" is sort of a misnomer since the "inertial frame" can also change relative to the "world frame" and I want to make sure people don't confuse the two.

velodyne_pointcloud/src/conversions/transform.cc
```diff
@@ -43,11 +43,11 @@ namespace velodyne_pointcloud
     srv_->setCallback (f);
     
     // subscribe to VelodyneScan packets using transform filter
-    velodyne_scan_.subscribe(node, "velodyne_packets", 10);
+    velodyne_scan_.subscribe(node, "velodyne_packets", 15);
```
JWhitleyAStuff: Is there a particular reason/rationale behind this change?

velodyne_pointcloud/src/conversions/transform.cc
```diff
@@ -43,11 +43,11 @@ namespace velodyne_pointcloud
...
     tf_filter_ =
       new tf::MessageFilter<velodyne_msgs::VelodyneScan>(velodyne_scan_,
                                                          listener_,
-                                                         config_.frame_id, 10);
+                                                         config_.fixed_frame_id, 100);
```
JWhitleyAStuff: Same here with 100 vs 10?

velodyne_pointcloud/cfg/TransformNode.cfg
mpitropov: I don't have a problem with making that change. However, I think is labelled as fixed_frame_id due to pcl_ros transformPointCloud calling the parameter fixed_frame with the description "fixed TF frame".

Since the inertial frame can move around, this makes me wonder if there is a potential bug here?

velodyne_pointcloud/src/conversions/transform.cc
mpitropov: I'll have to ask around about this one. This PR was before I started working with the team. The 100 seems excessive.

mpitropov:
Changes based on your comments:
1. The two numbers that were changed have been reverted. The team said it was changed to debug another node and not changed back.
2. I completely removed the added param because it made the other param not used.

Changes based on new commits:
1. Every time I saw "target" I changed it to "fixed". The transformPointCloud is given inPc_.pc->header.frame_id as the target frame and config_.frame_id as the fixed frame so what was on master didn't make much sense.
2. Every time I saw "odom" I changed it to "map". This is based on http://www.ros.org/reps/rep-0105.html which states "The pose of a mobile platform in the odom frame can drift over time, without any bounds.".

I think "frame_id" could be changed to be more specific like "fixed_frame_id" but I changed it back so that it doesn't break everyone's code.

velodyne_pointcloud/cfg/TransformNode.cfg
mpitropov: See new changes and top level comment

JWhitleyAStuff:
Thanks, @mpitropov ! One last thing - would you mind merging `master` back into your branch to fix CI? I switched CI providers because Travis has become SLLLOOOOOWWWW. Thanks.

mpitropov:
I merged in master. I also made one change to the test. I had it publishing map -> vehicle before but now there is one tf for map -> odom and another for odom -> vehicle to match REP 105.

image

PR #237, opened 2019-04-13
ROS2

klintan:
Draft PR, ported velodyne_msgs, which of course is super simple. Will do driver package next.

Wanted to keep a draft for feedback so that other in the community can follow the progress.

Disclaimer: I'm a junior C++ developer at best; meaning if the implementation is not up to par, I get that, no hard feelings. Other being, would love a lot of feedback on both small things as well as bigger things.

Edit: velodyne_driver is now implemented, it runs, but yet to test it fully. Missing tests as well. Needs some more work and feedback of course.


.gitignore
```diff
@@ -4,3 +4,7 @@
 *.user
 doc
 TAGS
+
+install/
+build/
+log/
```
JWhitleyAStuff: I'm assuming you added these because they are being generated in your `velodyne/` folder when you call `colcon build`. You can avoid this by calling `colcon build` from the base of your workspace (above `src/`). If you have other packages that you don't want built in your workspace, you can run `colcon build --packages-select <package names>`.

velodyne_driver/include/velodyne_driver/input.h
```diff
@@ -72,7 +75,7 @@ static uint16_t POSITION_PORT_NUMBER = 8308;  // default position port
 class Input
 {
 public:
-  Input(ros::NodeHandle private_nh, uint16_t port);
+  Input(rclcpp::Node * private_nh, uint16_t port);
```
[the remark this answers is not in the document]
JWhitleyAStuff: This applies to all other instances of raw pointers in this node for both NodeHandles and Messages. This makes sure there are no dangling pointers if the original owner of the pointer (which is most assuredly using a `std::shared_ptr`) goes out of scope and tries to clean up the allocated memory before this function is done working with the pointer.

velodyne_driver/package.xml
```diff
@@ -1,39 +1,52 @@
...
+    <version>1.5.2</version>
+    <description>
+        ROS device driver for Velodyne 3D LIDARs.
+    </description>
+    <maintainer email="jwhitley@autonomoustuff.com">Josh Whitley</maintainer>
+    <maintainer email="brice.rebsamen@gmail.com">Brice Rebsamen</maintainer>
+    <author>Jack O'Quin</author>
+    <author>Patrick Beeson</author>
+    <author>Michael Quinlan</author>
+    <author>Yaxin Liu </author>
```
JWhitleyAStuff: Please add yourself here.

velodyne_driver/package.xml
```diff
@@ -1,39 +1,52 @@
...
-  <depend>libpcap</depend>
-  <depend>nodelet</depend>
-  <depend>pluginlib</depend>
-  <depend>roscpp</depend>
-  <depend>tf</depend>
-  <depend>velodyne_msgs</depend>
+    
+    <buildtool_depend>ament_cmake</buildtool_depend>
+    
+    <build_depend>rclcpp</build_depend>
```
JWhitleyAStuff: Instead of repeating `<build_depend>` and `<exec_depend>`, you can use `<depend>` to do both.

velodyne_driver/src/driver/driver.cc
```diff
@@ -93,88 +91,80 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
...
-      ROS_ERROR_STREAM("unknown Velodyne LIDAR model: " << config_.model);
+      RCLCPP_ERROR(this->get_logger(), "unknown Velodyne LIDAR model: " + config_.model);
       packet_rate = 2600.0;
     }
   std::string deviceName(std::string("Velodyne ") + model_full_name);
 
-  private_nh.param("rpm", config_.rpm, 600.0);
-  ROS_INFO_STREAM(deviceName << " rotating at " << config_.rpm << " RPM");
+  this->get_parameter_or("rpm", config_.rpm, 600.0);
+  RCLCPP_INFO(this->get_logger(), deviceName + " rotating at " + std::to_string(config_.rpm) + " RPM");
```
JWhitleyAStuff: Especially in cases where multiple string concatenations are involved, please use the `sprintf` syntax for the logger macros. In this case, a temporary `std::string` object is created for each string concatenation. With the `sprintf` syntax, only a single temporary object is created for all replacements. e.g.

    RCLCPP_INFO(this->get_logger(), "%s rotating at %d RPM", deviceName, config_.rpm);

velodyne_driver/src/driver/velodyne_node.cc
```diff
@@ -35,22 +35,26 @@
...
+#include <rclcpp/rclcpp.hpp>
 #include "velodyne_driver/driver.h"
 
 int main(int argc, char** argv)
 {
-  ros::init(argc, argv, "velodyne_node");
-  ros::NodeHandle node;
-  ros::NodeHandle private_nh("~");
+  // Force flush of the stdout buffer.
+  setvbuf(stdout, NULL, _IONBF, BUFSIZ);
```
JWhitleyAStuff: Is there a reason this is necessary?

velodyne_driver/src/lib/CMakeLists.txt
```diff
@@ -1,12 +1,34 @@
...
+# libpcap provides no pkg-config or find_package module:
+set(libpcap_LIBRARIES -lpcap)
+
+include_directories(include
+  ${rclcpp_INCLUDE_DIRS}
+  ${rmw_implementation_INCLUDE_DIRS}
+  ${velodyne_msgs_INCLUDE_DIRS}
+)
+
 add_library(velodyne_input input.cc)
```
JWhitleyAStuff: This should be marked `SHARED` in ROS2 because of how the nodes use the libraries.

velodyne_msgs/CMakeLists.txt
```diff
@@ -1,16 +1,24 @@
...
+endif()
 
-add_message_files(
-  DIRECTORY msg
-  FILES
-  VelodynePacket.msg
-  VelodyneScan.msg
-)
-generate_messages(DEPENDENCIES std_msgs)
+find_package(tinyxml2 REQUIRED)
```
JWhitleyAStuff: This shouldn't be required.

velodyne_msgs/CMakeLists.txt
```diff
@@ -1,16 +1,24 @@
...
+rosidl_generate_interfaces(${PROJECT_NAME}
+   "msg/VelodynePacket.msg"
+   "msg/VelodyneScan.msg"
+   DEPENDENCIES std_msgs
+   DEPENDENCIES builtin_interfaces
 
-catkin_package(
-  CATKIN_DEPENDS message_runtime std_msgs
 )
+
```
JWhitleyAStuff: This should also have: 

    ament_export_dependencies(rosidl_default_runtime)

So the packages that depend on this messages one don't have to include this dependency.

velodyne_msgs/package.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0"?>
-<package format="2">
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
```
JWhitleyAStuff: If you're going to use package format 3 here, please do so with all of the other packages.

velodyne_msgs/package.xml
```diff
@@ -13,11 +14,21 @@
...
-  <buildtool_depend>catkin</buildtool_depend>
 
-  <build_depend>message_generation</build_depend>
+    <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <depend>std_msgs</depend>
+    <build_depend>builtin_interfaces</build_depend>
+    <build_depend>rosidl_default_generators</build_depend>
+    <build_depend>std_msgs</build_depend>
+    <build_depend>rosidl_default_generators</build_depend>
```
JWhitleyAStuff: This line is a duplicate. Others can be de-duplicated with `<depend>`.

JWhitleyAStuff:
@klintan - Also see my PR against your fork - https://github.com/klintan/velodyne/pull/1. It adds "working" CI (still fails due to dependency issues that will need to be resolved but the environment checks-out).

klintan:
Thank you so much for your patience and a lot of valuable feedback. Thanks for getting the circleci config going as well. 

I think I adressed most of your comments/feedback. I tested it at home as well with my VLP-16 and I got some data flowing :) So real happy about that 💯even though we are not at the finish line yet.

velodyne_msgs/package.xml
klintan: Will update the others.

velodyne_msgs/CMakeLists.txt
klintan: Something on my local mac that requires me to have it, but will keep it local and remove it from the repo.

velodyne_driver/src/lib/CMakeLists.txt
klintan: I got an error when making it `SHARED` but I'll see if I can get it fixed.

velodyne_driver/src/driver/velodyne_node.cc
klintan: Hm probably not, it's just in every ROS2 sample, so I just thought it was something that was best practice if you wanted to log to stdout in C++ ? But I can remove it,

velodyne_driver/include/velodyne_driver/input.h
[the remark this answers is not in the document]
klintan: Awesome thanks, makes a lot of sense.

JWhitleyAStuff:
Oops. I didn't mean to change this from a Draft PR. That's OK, though. Keep updating.

velodyne_driver/src/driver/CMakeLists.txt
```diff
@@ -1,26 +1,59 @@
...
+include_directories(include
+                    ${rclcpp_INCLUDE_DIRS}
+                    ${console_bridge_INCLUDE_DIRS}
+                    ${diagnostic_updater_INCLUDE_DIRS}
+                    ${rmw_implementation_INCLUDE_DIRS}
+                    ${velodyne_msgs_INCLUDE_DIRS}
+                    ${diagnostic_msgs_INCLUDE_DIRS}
+)
+#${rclcpp_components_INCLUDE_DIRS}
+
```
JWhitleyAStuff: Because of the `include_subdirectory()` directive in the top-level CMakeLists.txt, the items above this aren't necessary (they're already included in the top-level one). The `include_subdirectory()` in the top-level CMakeLists.txt tells CMake to add the CMakeLists.txt found in that subdirectory to the top-level one so they're processed as one file.

velodyne_driver/package.xml
```diff
@@ -1,39 +1,47 @@
...
+    
+    <buildtool_depend>ament_cmake</buildtool_depend>
+  
+    
+    <depend>rclcpp</depend>
+    <depend>rmw_implementation</depend>
+    <depend>diagnostic_updater</depend>
+    <depend>libpcap</depend>
+    <depend>pluginlib</depend>
+    <depend>console_bridge</depend>
```
JWhitleyAStuff: Change this to `libconsole-bridge-dev` and it will build in CI. See the first line at http://wiki.ros.org/console_bridge for explanation.

velodyne_driver/src/driver/driver.cc
```diff
@@ -93,88 +91,80 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
...
       // read data from live socket
-      input_.reset(new velodyne_driver::InputSocket(private_nh, udp_port));
+      input_.reset(new velodyne_driver::InputSocket(shared_from_this() , udp_port));
     }
-
+  
   // raw packet output topic
   output_ =
-    node.advertise<velodyne_msgs::VelodyneScan>("velodyne_packets", 10);
+    this->create_publisher<velodyne_msgs::msg::VelodyneScan>("~/velodyne_packets", 10);
```
JWhitleyAStuff: This shouldn't have the `~/` in front. Just having it as `velodyne_packets` makes it possible to namespace it properly.

velodyne_driver/src/driver/velodyne_node.cc
JWhitleyAStuff: I haven't found anything specifically stating that this is necessary. I guess it doesn't really hurt to leave it.

velodyne_driver/src/lib/CMakeLists.txt
```diff
@@ -1,12 +1,34 @@
...
+
+# libpcap provides no pkg-config or find_package module:
+set(libpcap_LIBRARIES -lpcap)
+
+include_directories(include
+  ${rclcpp_INCLUDE_DIRS}
+  ${rmw_implementation_INCLUDE_DIRS}
+  ${velodyne_msgs_INCLUDE_DIRS}
+)
+
```
JWhitleyAStuff: See my comment above about the `include_subdirectory()` directive.

velodyne_driver/src/lib/input.cc
```diff
@@ -136,9 +137,12 @@ namespace velodyne_driver
   }
 
   /** @brief Get one velodyne packet. */
-  int InputSocket::getPacket(velodyne_msgs::VelodynePacket *pkt, const double time_offset)
+  int InputSocket::getPacket(velodyne_msgs::msg::VelodynePacket *pkt, const double time_offset)
   {
```
JWhitleyAStuff: Missed one with the `std::shared_ptr<>`.

velodyne_driver/src/lib/input.cc
```diff
@@ -296,7 +301,7 @@ namespace velodyne_driver
   }
 
   /** @brief Get one velodyne packet. */
-  int InputPCAP::getPacket(velodyne_msgs::VelodynePacket *pkt, const double time_offset)
+  int InputPCAP::getPacket(velodyne_msgs::msg::VelodynePacket *pkt, const double time_offset)
```
JWhitleyAStuff: `std::shared_ptr<>`

klintan:
Thanks again, sorry for the slight delay, will try to complete the rest as soon as possible and address all outstanding feedback.

clalancette:
@klintan @JWhitleyAStuff Now that #251 is merged, I'll suggest that we close out this PR.  If we want to do a Crystal port, I think it will be easier to start from `dashing-devel` and use Crystal-only APIs (though I won't bother doing that work unless someone asks for it).  Does that make sense?

PR #239, opened 2019-04-30
Fix dead pointers in calibration.cc

joeygannon:
There are two places in calibration.cc where a temporary object is created, a pointer to that object is stored, the object goes out of scope, and then the pointer is dereferenced. This fix replaces the temporary objects with const references so the resulting pointers are valid.


JWhitleyAStuff:
@joeygannon Why not just make pointers to the objects directly? e.g.:

    max_intensity_node = &node[MAX_INTENSITY];

joeygannon:
@JWhitleyAStuff That'd be fine with me. I was trying to be as uninvasive as possible and respect the original author's style, but either way works.

JWhitleyAStuff:
I'm OK with doing away with the unnecessary ref.

JWhitleyAStuff:
Superseded by #265.

PR #243, opened 2019-05-20
driver: add enabled reconfigure param

dandedrick:
This will allow polling to be disabled when the device has been disabled
in some way. Specifically this is interesting when the rpm is set to 0
when not is use. This also adds a mask value for the time_offset so that
it won't be touched if it wasn't reconfigured.

### DIFF
--- a/velodyne_driver/cfg/VelodyneNode.cfg
+++ b/velodyne_driver/cfg/VelodyneNode.cfg
@@ -8,7 +8,8 @@ from dynamic_reconfigure.parameter_generator_catkin import *
 
 gen = ParameterGenerator()
 
-gen.add("time_offset", double_t,  0, "A manually calibrated offset (in seconds) to add to the timestamp before publication of a message.",
+gen.add("time_offset", double_t,  1, "A manually calibrated offset (in seconds) to add to the timestamp before publication of a message.",
         0.0, -1.0, 1.0)
+gen.add("enabled", bool_t, 2, "Switch to enable and disable lidar packet consumption", True);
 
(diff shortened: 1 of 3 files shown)


JWhitleyAStuff:
@dandedrick - I'm not a fan of defaulting the `time_offset` to 1. Can you find a way to do this with a default of 0?

dandedrick:
That's not the default value that's the level value. The default comes after the description and is still set to 0. See here: http://wiki.ros.org/dynamic_reconfigure/Tutorials/HowToWriteYourFirstCfgFile . the order of parameters is as follows:

 - name - a string which specifies the name under which this parameter should be stored

 - paramtype - defines the type of value stored, and can be any of int_t, double_t, str_t, or bool_t

 - level - A bitmask which will later be passed to the dynamic reconfigure callback. When the callback is called all of the level values for parameters that have been changed are ORed together and the resulting value is passed to the callback.

- description - string which describes the parameter

 - default - specifies the default value

 - min - specifies the min value (optional and does not apply to strings and bools)

 - max - specifies the max value (optional and does not apply to strings and bools)|

JWhitleyAStuff:
I'm still not quite understanding the use-case for this. You say when the device is disabled in some way - when would you disable the device but still want the driver to run?

dandedrick:
In order to keep from having to constantly stop and restart the driver. In order to save power and wear on the device when it's not actively needed it can have the rpm set to 0 and just quickly be disabled without having to stop the service. Then when it's needed the rpm can be set back to the desired value and the service be continued, rather than going through all the overhead of restarting the service.

PR #250, opened 2019-06-21
bug fix for row step

zhixy:
As "cloud.width" has not been assigned, it is 0 here, which makes "cloud.row_step" 0. Another option is moving the calculation of "cloud.row_step"  to function "setup".


JWhitleyAStuff:
@zhixy - Have you tested this on a sensor? If so, which one and what were the results?

zhixy:
Yes, I have test on Puck Hi-res with "organize_cloud=true", the row_step goes right, and the message can be converted to pcl pointcloud successfully. However, if row_step is 0, pcl pointcloud is wrong.

Issue #249, opened 2019-06-20
frame_id ignored and published with constant "velodyne"


omrirz:
**Please complete the following information:**
 - OS and Version: Ubuntu 16.04
 - ROS Version: Kinetic
 - Built from Source
 - Version: 16bc0c3

**Describe the bug**
velodyne/velodyne_pointcloud/src/conversions/convert.cc sets:
config_.target_frame = config_.fixed_frame = "velodyne";
and thus the frame_id value passed to launch files is being ignored.

**To Reproduce**
1.run roslaunch velodyne_pointcloud 32e_points.launch frame_id:="some_name"
2. run rostopic echo -p /velodyne_points/header
3. See fame_id is "velodyne"

**Expected behavior**
fame_id should be the argument passed to the launch file

JWhitleyAStuff:
Not sure how I missed this. Thanks for pointing it out! Any chance you could make a PR?

omrirz:
Yes. Will try to find time soon.

JWhitleyAStuff:
@omrirz - Can you please pull the branch `fix/fixed_frame_target_frame` and verify that this solves your issue? Thanks.

JWhitleyAStuff:
Solved by #253.

bmegli:
@JWhitleyAStuff , @omrirz 

This is still not working:
- in current master 96b39056bc16c5d3be9d32a709d4cd824189ef40
-in 4f042d92bccb9ba2c01356a3f8f586e8d3208c18 were it was supposed to be fixed 

It is working in 5e02d34cf8fe178844cb1d764a1cbd543596512f which I use

JWhitleyAStuff:
@bmegli - Is this still an issue with the current `master`? I haven't been able to look into it recently.

lesterlo:
@bmegli 

Yes, even on master branch (3d96504). The frame_id is still ignored.

The target_frame and fixed_frame are "velodyne" even the frame_id on the .launch file is modifed to other name.

pankit7:
> @bmegli - Is this still an issue with the current `master`? I haven't been able to look into it recently.

Yes, this is still an issue. As stated by Bartosz, it works in  [5e02d34](https://github.com/ros-drivers/velodyne/commit/5e02d34cf8fe178844cb1d764a1cbd543596512f) but not in the current master branch

JWhitleyAStuff:
[...]

PR #251, opened 2019-06-24
ROS 2 Dashing port

clalancette:
This PR is a fairly complete port of the velodyne driver to ROS 2 Dashing.  I started with the work of @klintan in #237, and did quite a bit of work to finish up the port; hence, this PR would supersede #237.  Note that it is currently targeted at "master" because I rebased the work of #237 on master; we'd probably want to make a "dashing-devel" branch instead.  I've tested the code on VLP16, and confirmed the following to work:

1.  Getting "raw" data packets out of a Velodyne and publishing to ROS 2 (the velodyne_driver)
1.  Taking the VelodyneScan and converting it to a pointcloud (the velodyne_pointcloud)
1.  Taking a published point cloud and get a laser scan line out of it (the velodyne_laserscan)

There are still some things that need to be worked on:
1.  Add the ability to change parameters on the fly (right now they can only be set at startup)
1.  Make the nodes composable (right now they are separate nodes; I'm waiting for https://github.com/ros2/rclcpp/pull/758 to land in Dashing before doing this)
1.  Port the integration tests
1.  Re-enable parts of the logging that I've disabled
1.  Test the "transform" velodyne_pointcloud node
1.  Do some static analysis on the code and fix things up
1.  Test on more Velodyne devices (though I don't have access to anything but a VLP16)
1.  Port to Crystal (this should be possible, but I didn't bother since it will make the PR bigger and I don't know if there is interest in it)
1.  Get the CI setup (I haven't played with this at all, it is probably broken)
1.  Get diagnostics working (this is blocked waiting on diagnostics to be ported to ROS 2)

However, before I go any further making changes, I'd like to discuss how to land this monstrosity.  This PR is probably too big to review in one go, so I'd be curious to hear from the maintainers (@JWhitleyAStuff , etc) how we want to proceed.  Any thoughts appreciated; thanks!


klintan:
Awesome work! thanks for taking this up, as you've gotten further than me, I just started taking it up again after doing some work on perception_pcl, I think we can continue from this PR maybe.

velodyne_pointcloud/src/conversions/transform.cc
```diff
@@ -18,110 +18,123 @@
...
 
 namespace velodyne_pointcloud
 {
   /** @brief Constructor. */
-  Transform::Transform(ros::NodeHandle node, ros::NodeHandle private_nh):
-    tf_prefix_(tf::getPrefixParam(private_nh)),
-    data_(new velodyne_rawdata::RawData),
-    first_rcfg_call(true)
+  Transform::Transform() :
+    rclcpp::Node("transform_node"),
```
klintan: See above regarding using the components approach in Dashing. Also I think we should keep naming consistent on all nodes, so maybe we rename the Convert-node to "convert_node" instead of "convert" only as it is now.

velodyne_pointcloud/src/conversions/transform_node.cc
```diff
@@ -12,20 +12,21 @@
...
-  ros::init(argc, argv, "transform_node");
-
-  // create conversion class, which subscribes to raw data
-  velodyne_pointcloud::Transform transform(ros::NodeHandle(),
-                                           ros::NodeHandle("~"));
+  rclcpp::init(argc, argv);
 
   // handle callbacks until shut down
-  ros::spin();
+  rclcpp::spin(std::make_shared<velodyne_pointcloud::Transform>());
```
klintan: If you change to the components approach you would need to do
``` 
rclcpp::NodeOptions options;
auto conv = std::make_shared<velodyne_pointcloud::Transform>(options);
```

velodyne_driver/include/velodyne_driver/driver.h
```diff
@@ -33,37 +33,33 @@
...
 
-class VelodyneDriver
+class VelodyneDriver final : public rclcpp::Node
 {
-public:
-  VelodyneDriver(ros::NodeHandle node,
-                 ros::NodeHandle private_nh);
-  ~VelodyneDriver() {}
+public :
+  VelodyneDriver();
```
klintan: I guess this could also be a component based node.

clalancette:
> Awesome work! thanks for taking this up, as you've gotten further than me, I just started taking it up again after doing some work on perception_pcl, I think we can continue from this PR maybe.

Excellent!  Yeah, I'm happy to do some more work and get this in, I just wanted to pause here so I can get a feeling of what direction to go in next.

JWhitleyAStuff:
@clalancette - Very nice work! I've re-targeted the PR to a new `dashing-devel` branch. Since I knew that the ROS2 port would be a nearly complete re-write, I wanted to suggest a couple of feature edits:

1. Make the "cut_at_specified_angle" feature the default with an angle of 2*pi radians. The current version defaults to a specific number of packets per revolution (`npackets`). Since the rotation rate of the sensor can vary, this causes over- and under-scanning in the generated point clouds. I think we should just do away with the `npackets` value and calculate some max estimated value for possible number of packets for pre-allocation purposes, then use the cut_at_specified_angle feature always.
2. Set "organize_point_cloud" defaulted to true.

Hopefully these aren't too much of a disruption but they should be fairly simple to implement. Thoughts?

velodyne_driver/mainpage.dox
```diff
@@ -3,7 +3,7 @@
 
 \htmlinclude manifest.html
 
-ROS device driver for Velodyne 3D LIDARs. 
+ROS device driver for Velodyne 3D LIDARs.
```
klintan: ROS2

velodyne_laserscan/src/node.cpp
```diff
@@ -30,20 +30,18 @@
...
-  ros::NodeHandle nh;
-  ros::NodeHandle nh_priv("~");
-
-  // create VelodyneLaserScan class
-  velodyne_laserscan::VelodyneLaserScan n(nh, nh_priv);
+  rclcpp::init(argc, argv);
 
   // handle callbacks until shut down
-  ros::spin();
+  rclcpp::spin(std::make_shared<velodyne_laserscan::VelodyneLaserScan>());
```
klintan: Not a biggie, but maybe here also follow how its done in the velodyne_pointcloud package 
```
auto laserscan = std::make_shared<velodyne_laserscan::VelodyneLaserScan>()
rclcpp::sping(laserscan
```
or similar.
(in addition, as mentioned on other this should probably be written as the new component thingy in Dashing and use NodeOptions)

clalancette:
> Hopefully these aren't too much of a disruption but they should be fairly simple to implement. Thoughts?

Both of those sound like fine changes to me, and I'm happy to do them.  However, I'll suggest that we use a separate PR to implement them.  Otherwise, those relatively small changes will be swamped by the size of the port, and thus it may be harder for users to see why the behavior on this branch is subtly different than on master.  Does that make sense?

JWhitleyAStuff:
@clalancette - Yes, absolutely. Thanks for the suggestion.

velodyne_driver/src/lib/input.cc
```diff
@@ -90,55 +97,57 @@ namespace velodyne_driver
...
+        RCLCPP_ERROR(private_nh->get_logger(), "Error binding to socket: %s", ::strerror(errno));
         return;
       }
-  
-    if (fcntl(sockfd_,F_SETFL, O_NONBLOCK|FASYNC) < 0)
+
+    if (::fcntl(sockfd_, F_SETFL, O_NONBLOCK|FASYNC) < 0)
       {
-        perror("non-block");
+        RCLCPP_ERROR(private_nh->get_logger(), "Error settign socket to non-blocking: %s", ::strerror(errno));
```
klintan: Typo, "setting"

velodyne_pointcloud/src/conversions/convert.cc
```diff
@@ -13,131 +13,146 @@
...
 #include <velodyne_pointcloud/organized_cloudXYZIR.h>
+#include <velodyne_pointcloud/pointcloudXYZIR.h>
+#include <velodyne_pointcloud/rawdata.h>
 
 namespace velodyne_pointcloud
 {
   /** @brief Constructor. */
-  Convert::Convert(ros::NodeHandle node, ros::NodeHandle private_nh):
-    data_(new velodyne_rawdata::RawData()), first_rcfg_call(true)
+  Convert::Convert() : rclcpp::Node("convert"),
```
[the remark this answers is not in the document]
clalancette: Yeah, that makes sense; I've done that now in 74c14b40496bb5406ca786b675da0e2f4b2c5db3, e2d9a4adb51dcf93cc7e6af5d4d4ffeaa93515f3, and 31b649f487a75c474096e9e348694288adb576ad

velodyne_pointcloud/src/conversions/convert.cc
```diff
@@ -13,131 +13,146 @@
...
     // publish the accumulated cloud message
-    diag_topic_->tick(scanMsg->header.stamp);
-    diagnostics_.update();
-    output_.publish(container_ptr_->finishCloud());
+    //diag_topic_->tick(scanMsg->header.stamp);
+    //diagnostics_.update();
+    output_->publish(container_ptr_->finishCloud());
   }
 
 } // namespace velodyne_pointcloud
```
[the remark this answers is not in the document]
clalancette: I didn't do this for two reasons:

1.  I'm waiting on ros2/rclcpp#758 to be merged and put into Dashing; that will auto-magically create both the components and a stand-alone node to launch things with.
1.  I wanted to not increase the size of this PR too much more.

velodyne_pointcloud/src/conversions/transform.cc
clalancette: I don't feel too strongly about this either way; I mostly left things as-is to keep the port smaller, rather than trying to improve everything at once.  I agree with you that the node names should be more consistent, but I also think it will be better to do that in a separate PR.

velodyne_pointcloud/src/conversions/convert.cc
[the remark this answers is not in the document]
klintan: ok awesome. Makes sense, I wasn't aware of that PR.

velodyne_pointcloud/src/conversions/transform.cc
klintan: I'm all good with that as well.

velodyne_driver/src/lib/input.cc
clalancette: Eek.  Good call; fixed in a0752c1a1d87792a5da43945d76caba3fe3b0742

velodyne_driver/src/lib/input.cc
```diff
@@ -90,55 +97,57 @@ namespace velodyne_driver
...
+        RCLCPP_ERROR(private_nh->get_logger(), "Error opening socket: %s", ::strerror(errno));
         return;
       }
-  
+
     sockaddr_in my_addr;                     // my address information
     memset(&my_addr, 0, sizeof(my_addr));    // initialize to zeros
     my_addr.sin_family = AF_INET;            // host byte order
-    my_addr.sin_port = htons(port);          // port in network byte order
+    my_addr.sin_port = ::htons(port);          // port in network byte order
```
klintan: Hm not sure this will work, since `htons` is a macro (at least on MacOS). I have been trying to compile it on Mac and it doesn't like this. (https://stackoverflow.com/questions/15002550/c-macros-and-namespaces)

JWhitleyAStuff: How about:

    #undef htons;
    my_addr.sin_port = htons(port);

Since this should only affect the code from this point to the end of this file during compilation, it shouldn't be an issue.

JWhitleyAStuff: It's failing CI as well on this item.

clalancette:
I've opened https://github.com/ros-drivers/velodyne/issues/252 to track some of the additional things we want to fix after we get the initial port in.  Feel free to add to that as necessary.

velodyne_driver/src/lib/input.cc
clalancette: Ah, I didn't realize it was a macro on some platforms.  Will revert back to just `htons`.

clalancette:
Nice.  Thanks to both of you, and a slight adjustment to the circleci config, CI is now passing.  I also think I've addressed all of the pending review comments (the ones that I haven't replied to yet basically all revolve around composability, which I think we should do later).  I also added commit 6b17e46 to try and fix the `fixed_frame`/`target_frame` bug that is present on master and reported in #249 ; we ran into this bug in our testing as well.

Are there any blockers to getting this merged?  I'm happy to make more changes, I just want to keep up the momentum.

PR #254, opened 2019-06-26
Dashing: Rename nodes.

### LINKED ISSUE #264: ROS 2 Dashing improvement: Rename the node names and file names to be consistent
RIght now, the node names aren't consistent (some have a "_node" suffix and some don't, some have a "velodyne_" prefix and some don't), and the file names don't necessarily match the node names.  Change all of this to be more consistent.

clalancette:
As discussed in https://github.com/ros-drivers/velodyne/pull/251,
we should make the default node names a lot more consistent.
Here, rename the nodes to all be
velodyne_{driver,convert,transform,laserscan}_node.  I also
renamed the "cloud_node.cc" file to be "convert_node.cc", so it
matches the rest of the code.  Finally, I removed the unused
nodelets.xml file from velodyne_pointcloud.

@JWhitleyAStuff @klintan FYI

Fixes #264


clalancette:
@JWhitleyAStuff Thanks for merging most of the other PRs.  Can we discuss this one and come to a conclusion here?  This is kind of holding me up, since if we take it, I have to change the node name in all of my parameter files.  I'm fine with doing that, but I just need a decision so we can move forward.  Thanks!

JWhitleyAStuff:
Sorry for the delay. It looks good to me and I'm in favor of making it more specific.

clalancette:
> Sorry for the delay. It looks good to me and I'm in favor of making it more specific.

Thank you!

PR #256, opened 2019-06-26
Dashing: Switch to using the cut angle by default.

### LINKED ISSUE #262: ROS 2 Dashing improvement: Use "cut_angle" feature by default
Make "cut_at_specified_angle" the default with an angle of 2*pi radians. The current version defaults to a specific number of packets per revolution (npackets). Since the rotation rate of the sensor can vary, this causes over- and under-scanning in the generated point clouds. I think we should just do away with the npackets value and calculate some max estimated value for possible number of packets for pre-allocation purposes, then use the cut_at_specified_angle feature always.

clalancette:
This ensures that we get a full scan rotation, not just
a somewhat variable number of packets.

@JWhitleyAStuff While I do like this default strategy better, in my testing the cut_angle feature didn't actually work for values less than 2*PI (or, at least, didn't work as I expected it to).  As far as I could tell, it never cut off anything; it always collected a full rotation.  And looking at the code in `poll`, this makes sense; there is no facility for "skipping" packets once we have passed the cut angle.  Am I misunderstanding the purpose of cut_angle?

Fixes #262

### DIFF
--- a/velodyne_driver/src/driver/driver.cc
+++ b/velodyne_driver/src/driver/driver.cc
@@ -76,7 +76,7 @@ VelodyneDriver::VelodyneDriver() : rclcpp::Node("velodyne_node")
   config_.model = this->declare_parameter("model", std::string("64E"));
   config_.rpm = this->declare_parameter("rpm", 600.0);
   std::string dump_file = this->declare_parameter("pcap", std::string(""));
-  double cut_angle = this->declare_parameter("cut_angle", -0.01);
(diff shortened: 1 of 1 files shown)


JWhitleyAStuff:
@clalancette - The `cut_angle` feature only works when it matches the settings of the Velodyne (max_angle - min_angle). It doesn't technically "cut" the data so much as align the number of produced points with the selected scan angle of the sensor.

PR #269, opened 2019-07-02
fix for #267, transform each packet

### LINKED ISSUE #267: A possible breaking change has been introduced in the TransformNodelet
I use the TransformNodelet to compensate for ego-motion. This gives me a more accurate point cloud in a global (odom or map) frame. All of this works fine when I use velodyne version 1.5.2.

However, when I am using master branch (https://github.com/ros-drivers/velodyne/tree/458bc7c72f607e7e99a93f5ad2ce79cc79b40fc1), the point clouds no longer seem to be corrected on the individual scan packet basis. It is as if the TransformNodelet does not transform individual scan packets, but only the entire scan message.

In version 1.5.2 it is clear that individual packets are transformed using the time stamp (https://github.com/ros-drivers/velodyne/blob/1.5.2/velodyne_pointcloud/src/conversions/transform.cc#L100). In master branch it seems that there is no such correction anymore (https://github.com/ros-drivers/velodyne/blob/master/velodyne_pointcloud/src/conversions/transform.cc#L134), which conforms to my experimental findings that the point clouds look less accurate.

spuetz:
This should fix #267.

### DIFF
--- a/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
@@ -113,13 +113,6 @@ public:
     cloud.width = config_.init_width;
     cloud.height = config_.init_height;
     cloud.is_dense = static_cast<uint8_t>(config_.is_dense);
-    if (config_.transform)
-    {
-      if (!computeTransformation(scan_msg->header.stamp))
-      {
-        ROS_ERROR_STREAM("Could not transform points!");
-      }
-    }
   }
 
   virtual void addPoint(float x, float y, float z, const uint16_t ring, const uint16_t azimuth, const float distance,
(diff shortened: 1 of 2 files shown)


spuetz:
The Hz-tests on the current master also fail on my machine.

Additionally, in the old code [here](https://github.com/ros-drivers/velodyne/blob/16bc0c391fdea3286c76ac3a220d0cbe7e614d86/velodyne_pointcloud/src/conversions/transform.cc#L113) the latest transformation in the tf buffer is used by using `ros::Time(0)`. This prevents the tf listener to wait for a transformation to become available. Should we also use the latest transformation here? Also we should think about using tf2.

JWhitleyAStuff:
@spuetz - the current `master` works fine on my machine, including the tests.

spuetz:
Looks like sometimes the Hz tests succeed and sometimes not. :/

JWhitleyAStuff:
From what I've seen, this is generally due to the load on the CI servers and the fact that the tests load data from a 64 which is pretty heavy. I gave the Hz tests a pretty wide berth (like 15 Hz) and that's about as wide as I can go without the tests becoming meaningless, IMO.

JWhitleyAStuff:
Yours ran fine on CI and my local machine this time.

RonaldEnsing:
> This should fix #265.

Could it be that you meant https://github.com/ros-drivers/velodyne/issues/267?

spuetz:
japp

PR #270, opened 2019-07-03
Dashing: Start removing node handle from input.cc

clalancette:
Ideally, the classes in [input.cc](https://github.com/ros-drivers/velodyne/blob/dashing-devel/velodyne_driver/src/lib/input.cc) would have no knowledge of ROS at all.  This would make them more independent and able to be unit tested a bit more easily.  This PR goes some way towards doing that by removing all of the `get_parameter` calls in input.cc, and instead passes those values in when the classes are constructed.  After this PR, the only remaining uses for the node handle inside of `input.cc` is for the logging; I'll consider how to deal with that better later.

### DIFF
--- a/velodyne_driver/src/lib/input.cc
+++ b/velodyne_driver/src/lib/input.cc
@@ -278,11 +278,11 @@ namespace velodyne_driver
     Input(private_nh, devip, port),
     packet_rate_(packet_rate),
     filename_(filename),
+    pcap_(nullptr),
     read_once_(read_once),
     read_fast_(read_fast),
     repeat_delay_(repeat_delay)
   {
-    pcap_ = NULL;
     empty_ = true;
 
     if (read_once_)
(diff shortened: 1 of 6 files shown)


clalancette:
@JWhitleyAStuff ping on this one; I'd like to get this in, since I have some more changes pending that build on top of this.  Thanks!

velodyne_driver/src/lib/input.cc
```diff
@@ -264,24 +265,26 @@ namespace velodyne_driver
...
+    repeat_delay_(repeat_delay)
   {
-    pcap_ = NULL;
     empty_ = true;
 
-    // get parameters using private node handle
-    private_nh->get_parameter("read_once", read_once_);
-    private_nh->get_parameter("read_fast", read_fast_);
-    private_nh->get_parameter("repeat_delay", repeat_delay_);
-
```
JWhitleyAStuff: I'm all for making this class ROS-independent but removing these `get_parameter` calls as well as the ones at lines 82/83 above removes the ability to pass these in as startup parameters. Can you please find a high-level way to gather these and pass them into the constructor? I know they're getting passed as dynamic reconfigure params but they should be taken as launch params as well and the launch params should override the DR params.

clalancette: It is somewhat subtle, but this is still configurable during startup.  In Dashing, the `declare_parameters` call actually declares the parameters *and* gets the current value.  The "current value" is defined as either what is passed in on the command-line during startup, or the default value to the `declare_parameters` call (as seen [here](https://github.com/ros-drivers/velodyne/pull/270/files#diff-5707fd8c8886aa445fb461efc9d17a2fR72)).

You can see this in action by cloning https://github.com/clalancette/declare , then building.  If you run without arguments, it will print 9.0:

```
ros2 run declare declare_node
```

If you run it pointing it to the params yaml file, you'll see 18.0 printed:

```
ros2 run declare declare_node __params:=src/declare/config/declare_params.yaml
```

Let me know if that assuages your concerns here.

clalancette:
Thanks!

PR #271, opened 2019-07-03
Dashing: Remove dead store from rawdata.cc

clalancette:
clang static analysis pointed out that distance is stored
to, but never read.  It looks like it is probably the result
of a bad merge, so just remove the unused variable.  Also
do a bit of style cleanup while we are in here.


JWhitleyAStuff:
@clalancette - Would you mind checking the same on `master`?

clalancette:
> @clalancette - Would you mind checking the same on `master`?

Will do!

JWhitleyAStuff:
@clalancette - Looking over this again, it is used in the code here too. Verify for me and I will revert this PR.

JWhitleyAStuff:
Nevermind - it gets redefined in this code.

PR #272, opened 2019-07-03
Remove a dead store from rawdata.cc.

clalancette:
It looks like probably the result of a bad merge; the distance
variable is never used in the outer scope, so just remove it.

### DIFF
--- a/velodyne_pointcloud/src/lib/rawdata.cc
+++ b/velodyne_pointcloud/src/lib/rawdata.cc
@@ -200,9 +200,6 @@ inline float SQR(float val) { return val*val; }
           continue;
         }
 
-        float distance = tmp.uint * calibration_.distance_resolution_m;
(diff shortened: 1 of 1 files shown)


velodyne_pointcloud/src/lib/rawdata.cc
```diff
@@ -200,9 +200,6 @@ inline float SQR(float val) { return val*val; }
           continue;
         }
 
-        float distance = tmp.uint * calibration_.distance_resolution_m;
-        distance += corrections.dist_correction;
-
```
JWhitleyAStuff: Looks like it does get used here in lines 215 and 239 below.

JWhitleyAStuff: Nevermind. I can't read.

PR #276, opened 2019-07-24
Dashing: Remove redundant storage of configuration data

clalancette:
This is a somewhat large and subjective PR, but the main goal here is to reduce the amount of duplicate storage of configuration parameters in the classes that make up the velodyne_pointcloud code.  The main benefits of doing so are:

1.  Helps with readability
1.  Makes it easier to implement on-the-fly parameter changes (which will come in a follow-on PR)
1.  Reduces memory footprint slightly

There is also a bunch of cleanup of code in this PR, along with the removal of some unused functions.  My suggestion for reviewing this is commit-by-commit, since each commit is fairly small and does one logical thing.  @JWhitleyAStuff if you'd prefer a couple of smaller PRs, I can do that too; just let me know.

Thanks!

### DIFF
--- a/velodyne_laserscan/src/velodyne_laserscan.cpp
+++ b/velodyne_laserscan/src/velodyne_laserscan.cpp
@@ -223,13 +223,7 @@ void VelodyneLaserScan::recvCallback(const sensor_msgs::msg::PointCloud2::Shared
         }
       else
         {
-          // TODO(clalancette): This can be simplified when RCLCPP_WARN_ONCE is available
-          static bool warned = false;
-          if (!warned)
-            {
-              RCLCPP_WARN(get_logger(), "VelodyneLaserScan: PointCloud2 fields in unexpected order. Using slower generic method.");
-              warned = true;
-            }
+          RCLCPP_WARN_ONCE(get_logger(), "VelodyneLaserScan: PointCloud2 fields in unexpected order. Using slower generic method.");
 
           if (offset_i >= 0)
             {
--- a/velodyne_laserscan/src/velodyne_laserscan.cpp
+++ b/velodyne_laserscan/src/velodyne_laserscan.cpp
@@ -223,7 +223,13 @@ void VelodyneLaserScan::recvCallback(const sensor_msgs::msg::PointCloud2::Shared
         }
       else
         {
-          RCLCPP_WARN(get_logger(), "VelodyneLaserScan: PointCloud2 fields in unexpected order. Using slower generic method.");
+          // TODO(clalancette): This can be simplified when RCLCPP_WARN_ONCE is available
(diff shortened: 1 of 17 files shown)


velodyne_laserscan/src/velodyne_laserscan.cpp
```diff
@@ -223,7 +223,13 @@ void VelodyneLaserScan::recvCallback(const sensor_msgs::msg::PointCloud2::Shared
         }
       else
         {
-          RCLCPP_WARN(get_logger(), "VelodyneLaserScan: PointCloud2 fields in unexpected order. Using slower generic method.");
+          // TODO(clalancette): This can be simplified when RCLCPP_WARN_ONCE is available
```
IanTheEngineer: Linked to https://github.com/ros2/rclcpp/issues/797

IanTheEngineer:
Looks great @clalancette! @JWhitleyAStuff have you had a chance to take a look at this?

velodyne_laserscan/src/velodyne_laserscan.cpp
clalancette: Actually, never mind.  We have `RCLCPP_WARN_ONCE`, we just don't have the `THROTTLE` versions.  Since we only need `WARN_ONCE` here, we are good.

clalancette:
@JWhitleyAStuff I know we are probably going to start concentrating on the other driver from Autoware, but in the meantime I would still like to get this merged (and potentially released into Dashing).  Any thoughts on this?

JWhitleyAStuff:
@clalancette I'm sorry I've been slow to respond. I'll review it today.

JWhitleyAStuff:
@clalancette - I've looked over everything and it seems pretty logical. It also seems like something we would want to do for the ROS1 Melodic driver. Have you tested this on a device yet?

clalancette:
> @clalancette I'm sorry I've been slow to respond. I'll review it today.

No worries!  Thanks for looking.

> @clalancette - I've looked over everything and it seems pretty logical. It also seems like something we would want to do for the ROS1 Melodic driver. Have you tested this on a device yet?

Yeah, I ran this on the VLP-16 I have (using Dashing).  There didn't seem to be much of a difference, which is what I was shooting for :).

I agree that this is something that would be nice to have in Melodic as well.  I suspect there will be a bunch of merge conflicts, though, both because of the port and because of some of the subsequent cleanup that I've done on the `dashing` branch.  What do you want to do here?

JWhitleyAStuff:
I'll go ahead and approve/merge this for now and we can implement it on Melodic later.

clalancette:
@JWhitleyAStuff Thanks for reviewing and merging!

PR #283, opened 2019-08-21
Feature/timestamp on scan start instead of end

### LINKED ISSUE #280: Adding an option to timestamp the velodyne_packets topic based on the first packet instead of the last packet in a scan
**Is your feature request related to a problem? Please describe.**
I would like the option to timestamp a velodyne_packets message based on the first packet instead of the last. It also seems to make more sense intuitively to me.

**Describe the solution you'd like**
Add a parameter option to set the velodyne_packets header stamp based on the first packet

**Describe alternatives you've considered**
n/a

**Additional context**
Implemented here:
https://github.com/kaarta/velodyne/tree/feature/timestampOnScanStartInsteadOfEnd

JWhitleyAStuff:
Closes #280.


JWhitleyAStuff:
@kaarta-SHanna - This change seems pretty reasonable to me and doesn't break any existing functionality. Have you tested this on a sensor? If so, which one?

kaarta-SHanna:
I have tested it on a vlp16, vlp32, and hdl32.

if we submit this and other enhancements, should we put an updated name/company on the copyright notice?

JWhitleyAStuff:
Yes, it's OK to add the current year and your/your company's name to the copyright if you provide contributions. As a rule-of-thumb, it's usually only if you provide significant contributions to a file but it's truly up to you.

Issue #213, opened 2019-02-01
DriverNodelet's Poll Timeout Handling


jimmyw404:
In the DriverNodelet::devicePoll() function if the VelodyneDriver::poll() call fails, the devicePoll thread will exit. This happens if there is a network disconnect, the puck's power cycles etc.

This was requiring me to restart the velodyne Driver nodelet whenever I had a timeout.

Is the intent that there is some other mechanism to restart the DriverNodelet when it is detected it exits? What is the recommended way to do this?

For now I've commented out the break; line. My DriverNodelet::devicePoll() function looks like this:

```
/** @brief Device poll thread main loop. */
void DriverNodelet::devicePoll()
{
  while(ros::ok())
    {
      // poll device until end of file
      running_ = dvr_->poll();
      if (!running_)
      {

      	ROS_ERROR("W: DriverNodelet:devicePoll poll failed, who cares?");
//        break;
      }
    }
  running_ = false;
}
```

And now it persists through disconnects and works great, but I'm guessing this is the inappropriate way to do this.

JWhitleyAStuff:
@jimmyw404 - Check the branch `maint/poll_timeout_handling`. Let me know if it works and I'll merge it if it does.

jimmyw404:
@JWhitleyAStuff Thank you for that code change. It will work just fine.

PR #284, opened 2019-08-25
when device poll timeouts, do not stop node

dseifert:
As a follow-up to #213 and the implementation in #216, I added support for not exiting the driver on poll timeouts to the node as well. I assume there is no specific reason why the node should behave differently than the nodelet here.


dseifert:
@JWhitleyAStuff, yes, on a VLP-16. However change in behavior is (intented) only when there is no live sensor (anymore) ;-)

Issue #242, opened 2019-05-14
Introduce phase_lock_angle param to control msg header timestamp


mpitropov:
**Is your feature request related to a problem? Please describe.**
Within the driver the packet message is published using the timestamp of the last packet. This means that the published pointcloud has the timestamp of the last packet. Also the transformed pountcloud is corrected to the timestamp of the last packet.
https://github.com/ros-drivers/velodyne/blob/7ecdf17173a609234df4e61cdd443748ccf7ab8f/velodyne_driver/src/driver/driver.cc#L243
This creates a problem with time sync where someone has a PPS signal triggering other sensors and also has a phase lock angle for their lidar. The lidar will have a delayed timestamp compared to other sensors.

**Describe the solution you'd like**
I think I'll have to create a param called phase_lock_angle which should be the same angle you have set in your lidar configuration. Adding to the current cut_angle loops seems to be the way to go although the double while loop is looking a bit large. When the packet angle passes the phase_lock_angle I will store the packet timestamp to be used for the published message header.

**Describe alternatives you've considered**
I have the cut angle set to behind the car and the phase lock set to in front of the car so in theory I would be able to just grab the timestamp of the center packet. Other people may have different setups though.

**Additional context**
![IMG_20190514_112242](https://user-images.githubusercontent.com/3324075/57710387-dd084c00-763a-11e9-9e15-eb9473c2282d.jpg)

JWhitleyAStuff:
I agree with the general concept. Would you be willing/able to put in a Pull Request with this implemented?

PR #285, opened 2019-08-27
Adding more precise scan separation feature

n-patiphon:
Hi all.
I'm working on some mapping projects at MapIV, Inc. (TierIV's subsidiary company). During the projects, I found a way to improve the current Velodyne ROS driver, so my team and I think it might be worth sharing here. If I'm missing something or I got something wrong, I would be more than happy to hear from you all :)

It seems to me that the current implementation only allows creating one full scan from a set of complete packets, say 76 packets/scan or 77 packets/scan. But sometimes, one full scan is actually equivalent to some complete packets and a fraction of it, for example, 76.3 packets/scan. In this case, rounding the number of packets/scan up or down will result in redundant points or missing points in a scan, respectively (see the figures below).

**Redundant points**
imperfect2
**Missing points**
imperfect1

So, what I did was basically modifying the driver to separate each scan more precisely. The separation mainly relies on azimuth data each point was taken (estimated using the azimuth of the packet and the firing timing). If the azimuth differences of three consecutive points have different signs, it is assumed that the last point belongs to a new scan. Therefore, it is stored in a residual point cloud waiting to be merged to the next scan. The code for this can be found [here](https://github.com/ros-drivers/velodyne/compare/master...MapIV:feature/precise_scan_separation?expand=1#diff-70aae7a710580a1d71ad71c8e11bedf6R503-R522).

Also, it is important to ensure that the number of packets in velodyne_msgs::VelodyneScan is more than the final exact packets/scan. For example, if the final number of packets/scan is 76.3, velodyne_msgs::VelodyneScan needs to contain 77 packets in total. The code for this part can be found [here](https://github.com/ros-drivers/velodyne/compare/master...MapIV:feature/precise_scan_separation?expand=1#diff-f19c2683e7974bb6c1e3e67865dccab2R142-R187).

Lastly, I also use the estimated timestamp of the last point in a scan as the scan timestamp instead of using the last packet timestamp as in the current implementation or the first packet as mentioned [here](https://github.com/ros-drivers/velodyne/issues/280).

**Video comparing current driver and the modified driver**
[IMAGE ALT TEXT](http://www.youtube.com/watch?v=YOCg2783kN4 "Modified Velodyne ROS driver")

**Limitations**
- Only works with VLP16 and VLP16-Hires
- Only works with unsorted Point Cloud
- Might be slower than the current implementation as it includes more processing

**Related discussions**
- https://github.com/ros-drivers/velodyne/issues/280
- https://github.com/ros-drivers/velodyne/issues/242
- https://github.com/ros-drivers/velodyne/issues/231

### DIFF
--- a/velodyne_pointcloud/include/velodyne_pointcloud/convert.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/convert.h
@@ -88,6 +88,12 @@ class Convert
     Config config_;
     bool first_rcfg_call;
 
+    // Additional parameters for checking azimuth differences between consecutive data blocks
+    velodyne_msgs::VelodyneScanPtr adjusted_vel_msg_ptr {new velodyne_msgs::VelodyneScan};
+    int last_azimuth_;
+    int current_azimuth_diff_;
+    int prev_azimuth_diff_;
+
   // diagnostics updater
   diagnostic_updater::Updater diagnostics_;
   double diag_min_freq_;
--- a/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
@@ -113,6 +113,7 @@ public:
     cloud.width = config_.init_width;
     cloud.height = config_.init_height;
     cloud.is_dense = static_cast<uint8_t>(config_.is_dense);
+    is_newscan = false;
   }
 
   virtual void addPoint(float x, float y, float z, const uint16_t ring, const uint16_t azimuth, const float distance,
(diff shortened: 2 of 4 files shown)


n-patiphon:
Hi @JWhitleyAStuff ,
Thank you for your suggestions. I already fixed the linting errors. 

As for the `cut_angle` parameter, I am aware of this implementation. But, as far as I understand this option still separates scans at a packet-level because it uses azimuth from a packet header. This header azimuth is the angle of the first return in the first firing sequence of that packet. So, it will still have the missing/redundant points problem I mentioned earlier. On the other hand, the implementation in this PR separates scans at a point-level because it uses interpolated azimuth of each point to decide whether they are in the same scan or not.

Besides, I went and rechecked the current master. As you suggested, I hard-coded the  `cut_angle`  to 2PI and ran the driver. These are the results.

master
pr

As you can see, the current master created a vertical dense patch shown in the red circle while this PR implementation did not.

JWhitleyAStuff:
@n-patiphon - You mentioned that this "might be slower" due to more processing. Can you possibly do a simple CPU load test while running the node with and without these modifications and report the results? Also, what else would be required to apply this logic to the other models?

JWhitleyAStuff:
@n-patiphon - There are also conflicts with the current `master`. Please rebase.

YoshuaNava:
@n-patiphon This is a very nice contribution. Thank you.

Do you have any updated about the CPU usage and impact on measurements timing?

spuetz:
Hi, I just looked into the code. I would suggest to just add another container and filtering on the phi angle (with respect to the spherical coordinate system). The container could buffer the extra points and add them - if necessary - during the setup method. I think this should be a much more elegant way and it should not consume more processing time / power. Or, am I missing something? Cheers

JWhitleyWork:
@spuetz or @YoshuaNava Would either of you be interested in continuing the implementation of this feature? I'm not sure that MapIV is still a thing (my contact at TierIV says he hasn't heard anything about them for quite some time).

spuetz:
Should be easy, I could do that in a couple of weeks. But I think I'll restart with the approach I suggested above.

JWhitleyWork:
> Should be easy, I could do that in a couple of weeks. But I think I'll restart with the approach I suggested above.

Once you have a new PR ready, I'll close this one.

spuetz:
I wait with this for #347 and #335 to be merged.

happengor:
This modification is great, with this modification i get a predictable motion distortion. With original implementation i get an unpredictable motion distortion, which is resulted from the changing of beginning time of scan. I hope this modification could soon be available by VLP-32C.

JWhitleyWork:
@n-patiphon @spuetz Pings all around.

PR #286, opened 2019-08-27
use correct node handles and node names for diagnostic_updater::Updater

dseifert:
`diagnostic_updater::Updater` by default is using ros::NodeHandle() and the node's name via ros::this_node::getName(). However this is not ideal for nodelets, as both refer to the nodelet manager instead of the nodelet. For example running a single velodyne driver nodelet this can be somewhat ignored, however running two (or more) will cause issues as it can not be differentiated which nodelet is sending what diagnostic output (or even how many nodelets are sending).

`diagnostic_updater::Updater` allows to pass the proper node handles and node name. This is implemented here, in order to be able to run multiple nodelets of the same type in the same nodelet manager.


dseifert:
@JWhitleyAStuff this was tested with live sensors (VLP-16).

PR #289, opened 2019-09-30
Dashing: Setup for components

### LINKED ISSUE #258: ROS 2 Dashing improvement: Make composable nodes
Right now, all of the nodes (driver, pointcloud, laserscan) are separate, meaning a lot of data is being copied back and forth. If we made them composable, they could share pointers, greatly improving their performance. We're currently waiting on ros2/rclcpp#758 to land and to make it into Dashing before making them composable.

clalancette:
This PR contains the changes necessary to allow the various parts of the Velodyne driver to be used as components, and thus to be composed into the same process.

This involves:
* Changing the build so that the node files are built as shared objects.
* Adding an `RCLCPP_REGISTER_COMPONENT` macro call at the end of the Node cpp files.
* Calling `rclcpp_component_register_nodes` in the build (note that this *must* be in the top-level CMakeLists due to a bug in Dashing).
* Adding `NodeOptions` to the node constructors.

This fixes #258 .  @JWhitleyAStuff Thanks in advance for any thoughts/review.


JWhitleyAStuff:
Nice work! Have you tested some or all of these in a multi-process configuration?

clalancette:
> Nice work! Have you tested some or all of these in a multi-process configuration?

Yep!  I ran all three of `velodyne_driver`, `velodyne_pointcloud`, and `velodyne_laserscan` in a single composed container, and everything worked out well with it.

velodyne_pointcloud/include/velodyne_pointcloud/transform.h
```diff
@@ -61,7 +61,7 @@ namespace velodyne_pointcloud
 class Transform final : public rclcpp::Node
 {
 public:
-  Transform();
+  explicit Transform(const rclcpp::NodeOptions& options);
   ~Transform()
```
JWhitleyAStuff: Can you please remove the explicit constructor? With this here, the compiler won't generate copy or move operations and it isn't doing anything.

clalancette: I don't care hugely about this, so I'll go either way here.  But in my (limited) testing here, having an empty destructor doesn't seem to disable the copy constructor.  I'm less certain about the move constructor (it's a little harder to write a straightforward example that shows it).

But actually, I'd argue that we explicitly don't *want* a copy or move constructor.  There's enough resources in here that I'm not at all certain we'd do the right thing on copy.  From that perspective, it might be better to introduce a `NO_COPY_NO_MOVE_NO_ASSIGN` macro (like [here](https://github.com/ros-drivers/phidgets_drivers/blob/10c1ca6ab712398c6c398b85a175ae20ae3e442e/phidgets_api/include/phidgets_api/phidget22.hpp#L38)), and explicitly not allow those operations.

What do you think?

JWhitleyAStuff: I'm OK with removing the copy and move operators. Here is the proper way:

```
Transform(Transform && t) = delete;
Transform &operator=(Transform && t) = delete;
Transform(const Transform &) = delete;
Transform &operator=(const Transform &) = delete;
```

clalancette: Great, thanks.  Fixed in 3ec6c52

JWhitleyAStuff:
Thanks! One last thing: Could you possibly create a composed node and launch file that includes the whole pipeline? That will make it much easier for users to transition from the ROS1 version.

clalancette:
> Thanks! One last thing: Could you possibly create a composed node and launch file that includes the whole pipeline? That will make it much easier for users to transition from the ROS1 version.

Sounds good.  Which package would you like it to live in?

JWhitleyAStuff:
Since the high-level launch files live in `_pointcloud`, let's put it there.

clalancette:
> Since the high-level launch files live in `_pointcloud`, let's put it there.

I took a brief look at doing this, and I think that having overarching launch files in `velodyne_pointcloud` is a mistake.  The basic reason is that the three packages in here (ignoring `velodyne_msgs`) have no direct dependencies on each other.  For all intents and purposes, they are siblings.  Thus, if someone does `apt-get install ros-dashing-velodyne-pointcloud`, they will get a pile of launch files that can't work.  Instead, I'll propose that we have a launch file for each of the siblings that just shows how to launch that sibling (in both composed and non-composed modes).  Then we add a launch file to the top-level `velodyne` package that shows launching all three together (in both composed and non-composed modes).

Which brings me to the second point.  I'm happy to do the above, but it is a lot of additional changes.  So my suggestion is that we merge this PR as-is, and I'll do that cleanup as a follow-up PR (likely early next week).

What do you think of this proposal?

JWhitleyAStuff:
@clalancette - This hasn't been done a ton as-of-yet but I want to make sure we're following suggested best practices: The recommended way to export components as shared libraries (see [this post](https://discourse.ros.org/t/ament-best-practice-for-sharing-libraries/3602)) is to add the following to CMakeLists.txt (just using `velodyne_driver` as an example):

```
ament_export_interfaces(export_velodyne_driver HAS_LIBRARY_TARGET)
ament_export_libraries(velodyne_driver)
install(
  TARGETS velodyne_driver
  EXPORT export_velodyne_driver
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION bin
)
```

And, when using the component downstream, users would link against `velodyne_driver::VelodyneDriver` (namespaced interface name) instead of `velodyne_driver` (library name). Would you mind implementing this on the modified packages?

clalancette:
For what it is worth, the style that I've done here (and in a bunch of other packages that I've made composable) comes directly from the Dashing demos:

https://github.com/ros2/demos/blob/dashing/composition/CMakeLists.txt

I think that's because we haven't totally gotten all of the issues around the "modern", exported style of linking to work; see https://github.com/ament/ament_cmake/pull/178 , for instance.

So I can look into it if you insist, but I'd rather get this PR in as-is and then later on take another crack at this (I have some changes queued up that will be Eloquent-only anyway).  I'm happy to open an issue to track it.

Thoughts?

clalancette:
Thanks for merging!  I'll look into updating the launch files in the next couple of days.

PR #291, opened 2019-10-03
Dashing: Properly setup the row_step.

clalancette:
We were setting it up too early, so that it was always zero.
Instead move it into the setup once we know the width of
the cloud.

@JWhitleyAStuff FYI

### DIFF
--- a/velodyne_pointcloud/tests/CMakeLists.txt
+++ b/velodyne_pointcloud/tests/CMakeLists.txt
@@ -1,52 +1,53 @@
-### Unit tests
-#
-#   Only configured when CATKIN_ENABLE_TESTING is true.
-
 # these dependencies are only needed for unit testing
-find_package(roslaunch REQUIRED)
-find_package(rostest REQUIRED)
-find_package(tf2_ros REQUIRED)
+find_package(ament_cmake_gtest REQUIRED)
+
+ament_add_gtest(test_row_step test_row_step.cpp)
+ament_target_dependencies(test_row_step
+  rclcpp
+  tf2
+  velodyne_msgs
+)
 
 # C++ gtests
-catkin_add_gtest(test_calibration test_calibration.cpp)
-add_dependencies(test_calibration ${catkin_EXPORTED_TARGETS})
-target_link_libraries(test_calibration velodyne_rawdata ${catkin_LIBRARIES})
+# catkin_add_gtest(test_calibration test_calibration.cpp)
+# add_dependencies(test_calibration ${catkin_EXPORTED_TARGETS})
+# target_link_libraries(test_calibration velodyne_rawdata ${catkin_LIBRARIES})
 
-# Download packet capture (PCAP) files containing test data.
-# Store them in devel-space, so rostest can easily find them.
-catkin_download_test_data(
-  ${PROJECT_NAME}_tests_class.pcap
-  http://download.ros.org/data/velodyne/class.pcap
-  DESTINATION ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_SHARE_DESTINATION}/tests
-  MD5 65808d25772101358a3719b451b3d015)
-catkin_download_test_data(
-  ${PROJECT_NAME}_tests_32e.pcap
-  http://download.ros.org/data/velodyne/32e.pcap
-  DESTINATION ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_SHARE_DESTINATION}/tests
-  MD5 e41d02aac34f0967c03a5597e1d554a9)
-catkin_download_test_data(
-  ${PROJECT_NAME}_tests_64e_s2.1-300-sztaki.pcap
-  http://download.ros.org/data/velodyne/64e_s2.1-300-sztaki.pcap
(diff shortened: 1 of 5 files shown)


JWhitleyAStuff:
@clalancette - Could you possibly write a test for this?

clalancette:
> @clalancette - Could you possibly write a test for this?

Good idea.  Done in 252c3cc

velodyne_pointcloud/tests/CMakeLists.txt
```diff
@@ -1,52 +1,53 @@
...
-
 # these dependencies are only needed for unit testing
-find_package(roslaunch REQUIRED)
-find_package(rostest REQUIRED)
-find_package(tf2_ros REQUIRED)
+find_package(ament_cmake_gtest REQUIRED)
+
+ament_add_gtest(test_row_step test_row_step.cpp)
+ament_target_dependencies(test_row_step
+  rclcpp
```
JWhitleyAStuff: Your test isn't using `rclcpp` so this can be removed.

JWhitleyAStuff:
Thanks for adding the test! I just had one comment on the implementation. It's also my goal to add `ament_lint_auto` to this package sometime soon (once I get time to do so).

velodyne_pointcloud/tests/CMakeLists.txt
clalancette: Ah yes, but `datacontainerbase.h` *is* using rclcpp (for time), and since this test is using it in header-only mode, we still need the dependency.

clalancette:
Thanks for reviewing and merging!

Neel1302:
Hi @clalancette and @JWhitleyWork,

Thanks for this fix.

I understand your fix for the `row_step` bug works for ROS2. I replicated the changes for ROS melodic and the master branch (the melodic devel or master branches do not seem to have this fix) and it seems to fail (I am using a VLP32e).

image

I have not changed the `init_width` and `init_height` parameters.

Another side effect after applying the change was that a lot of the data includes zeros:

image

Do you have any idea as to what maybe causing this?

Thanks,
Neel

JWhitleyWork:
@Neel1302 Do you mean VLP-32C (Ultra Puck) or HDL-32E?
Do you have a branch that we could look at? It is difficult to say if a typo was made or something distro-specific is causing the bug without looking at the changes you have made.

Neel1302:
Hi @JWhitleyWork,

I indeed have a Velodyne HDL-32E.

Please refer to [Neel1302/velodyne](https://github.com/Neel1302/velodyne) and specifically this file: [datacontainerbase.h](https://github.com/Neel1302/velodyne/blob/master/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h).

This file contains the modification I made as per #291 on the master branch (not the dashing branch)

Would be great if you could have a look.

Thanks,
Neel

JWhitleyWork:
@Neel1302 I noticed that line 80 in the Dashing version is:

```
cloud.row_step = cloud.width * cloud.point_step;
```

While the same line in the Noetic version is:

```
cloud.row_step = init_width * cloud.point_step;
```

I doubt this is the issue since `cloud.width` is set to `config_.init_width` which is set to `init_width` in the initialization list but it's worth testing.

clalancette:
> @clalancette I'm not seeing anything out-of-place here. What are we missing?

It's been a while since I looked at this, so I don't remember all of the details.  But I do remember that it was important for [these](https://github.com/ros-drivers/velodyne/blob/e154fd275f52ef5cae60a4a78f31ec5582afbc11/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.hpp#L122-L126) lines of code to be in there (and be *before* the resize).  Also, I think we need [this line](https://github.com/ros-drivers/velodyne/blob/e154fd275f52ef5cae60a4a78f31ec5582afbc11/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.hpp#L129) in the organized_cloud case.  Maybe that is what is missing?

Neel1302:
Hi @clalancette,
I agree that those lines are important. I think, however, you pointed out earlier that those lines had to be moved as in:

> We were setting it up too early, so that it was always zero.
> Instead move it into the setup once we know the width of
> the cloud.
> 
> Signed-off-by: Chris Lalancette [clalancette@openrobotics.org](mailto:clalancette@openrobotics.org)
> 
> @JWhitleyAStuff FYI

and the changes you made to the dashing branch reflect that:
image

Since this change has not been pushed to the melodic branch, I decided to add it in [Neel1302/velodyne](https://github.com/Neel1302/velodyne) and specifically in this file: [datacontainerbase.h](https://github.com/Neel1302/velodyne/blob/master/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h).

However after testing with this, my results above indicate that row_step is still zero - the focus of this issue.

I would appreciate if you could have a look and guide me. I would be willing to test any potential modifications, if necessary.

Thanks,
Neel

clalancette:
> However after testing with this, my results above indicate that row_step is still zero - the focus of this issue.
> 
> I would appreciate if you could have a look and guide me. I would be willing to test any potential modifications, if necessary.

Sorry, I don't know what's going on here.  My next suggestion is to go in with the debugger, put a watch on the row_step variable, and see what is modifying it (if anything).

PR #293, opened 2019-10-10
Dashing: Add in example launch files.

clalancette:

As promised in https://github.com/ros-drivers/velodyne/pull/289#issuecomment-539635729 , here is a PR that adds in example launch files for all of the parts of this driver.  While we are in here, we also add in README.md for each of the packages to give at least basic documentation on each node.  Finally, I also renamed some of the binaries so that they are a lot more consistent.  This will be slightly annoying to downstream users, but since we haven't released this yet, I think it is still OK to do.  Any thoughts on this are appreciated.


clalancette:
@JWhitleyAStuff Friendly ping; any thoughts on these example launch files?

velodyne_laserscan/src/velodyne_laserscan.cpp
```diff
@@ -183,25 +183,25 @@ void VelodyneLaserScan::recvCallback(const sensor_msgs::msg::PointCloud2::Shared
   // Construct LaserScan message
   if ((offset_x >= 0) && (offset_y >= 0) && (offset_r >= 0))
     {
-      const float RESOLUTION = std::abs(resolution_);
-      const size_t SIZE = 2.0 * M_PI / RESOLUTION;
+      const float resolution = std::abs(resolution_);
```
JWhitleyAStuff: Was there a reason for doing this? The "standard" is here from the Google C++ Style Guide: https://google.github.io/styleguide/cppguide.html#Constant_Names

velodyne_laserscan/src/velodyne_laserscan.cpp
```diff
@@ -183,25 +183,25 @@ void VelodyneLaserScan::recvCallback(const sensor_msgs::msg::PointCloud2::Shared
...
+      const size_t size = 2.0 * M_PI / resolution;
```
JWhitleyAStuff: Same question here.

JWhitleyAStuff:
Thanks for working on this! However, I don't know that we need both composed and non-composed versions of each of the launch files - we can probably do with just the non-composed versions when they're launching a single node and leave the one composed version for combining multiple nodes. Thoughts?

velodyne_laserscan/src/velodyne_laserscan.cpp
clalancette: So, they aren't macros, so they definitely shouldn't be all upper-case.  I could be convinced to use the `kResolution` form, but even then they aren't really constant; the resolution can be changed via parameter, so these are still variables in my mind.

That being said, I won't insist on it, so if you want to go back to the old form I can do this.

clalancette:
> Thanks for working on this! However, I don't know that we need both composed and non-composed versions of each of the launch files - we can probably do with just the non-composed versions when they're launching a single node and leave the one composed version for combining multiple nodes. Thoughts?

I think it is somewhat nice to have both the composed and non-composed versions for everything.  For instance, the user may want to compose the `velodyne_pointcloud` with their own home-grown node.  Having the example right in front of them in the same place means it is easier to copy-n-paste without having to go into a different package.

But I don't feel too strongly about it, so if you still think we should get rid of the composed examples for all but the `velodyne` package, I can do that.

velodyne_laserscan/src/velodyne_laserscan.cpp
JWhitleyAStuff: I would prefer the `kResolution` form just to denote that they are constant within the context of their scope, if you wouldn't mind.

clalancette: Done in 8ae7f648d91a27c4599c99058237fe7ca4e44dbf

clalancette:
Thank you!

PR #299, opened 2019-10-30
Check for intra-process subscribers

pjreed:
When running as a composed node in a container, the usual `get_subscription_count()` method may return 0; it's necessary to also check `get_intra_process_subscription_count()`.

Distribution Statement A; OPSEC #2893

Signed-off-by: P. J. Reed <preed@swri.org>


pjreed:
Yep, I have tested this on a VLP-16.

Issue #298, opened 2019-10-30
velodyne_convert_node doesn't work in ROS 2 Dashing


pjreed:
**Please complete the following information:**
 - OS and Version: Ubuntu 18.04
 - ROS Version: Dashing
 - Built from Source or Downloaded from Official Repository: Built from source
 - Version: e84f136fdc0132798da8e73fcd2906ecc57a064c

**Describe the bug**
The `velodyne_convert_node` seems to be performing very badly in ROS 2 Dashing.  I've got a vehicle with a VLP-16 that works normally using the ROS 1 Melodic driver; in Dashing, `velodyne_driver_node` publishes `/velodyne_packets` at a steady 10 Hz, but `velodyne_convert_node` publishes `/velodyne_points` sporadically when it first starts and eventually stops publishing at all.  Its memory usage also climbs very quickly; after running for about 10 minutes, the `velodyne_convert_node` process is consuming about 8 GB of memory on my system.

**To Reproduce**
1. Tweak `velodyne/launch/velodyne-all-nodes-VLP16-launch.py` as appropriate for your Velodyne
2. Launch it

**Expected behavior**
I expected `/velodyne_points` to be published at 10 Hz.

**Additional context**
This problem seems to happen both if the nodes are composed into a container and if they are run standalone -- however, note that if the container is using intra-process communication, it may not publish anything at all because the subscriber count may be incorrect.

pjreed:
As an addendum: when I do `rostopic echo /velodyne_points`, the size of the pointcloud is constantly growing.  If `velodyne_convert_node`'s `organize_cloud` param is `true`, the width is 16 but the height is constantly growing; if it's `false`, height is 1 but width is constantly growing.  Either way, it seems likely that it is re-using a data structure and not clearing out the old data with every new cloud it processes.

pjreed:
Fixed by #307

PR #307, opened 2019-11-06
Dashing fixes

clalancette:
This PR has a bunch of fixes that I needed to make the current Dashing actually work with my VLP-16.  With all of these in place, I'm now able to make this produce pointclouds as before.  @JWhitleyAStuff  please take a look, thanks!

### DIFF
--- a/velodyne_driver/include/velodyne_driver/driver.hpp
+++ b/velodyne_driver/include/velodyne_driver/driver.hpp
@@ -74,7 +74,7 @@ private:
   }
   config_;
 
-  std::shared_ptr<Input> input_;
+  std::unique_ptr<Input> input_;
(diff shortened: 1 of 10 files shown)


JWhitleyAStuff:
@clalancette - Everything in this MR looks fine. Do you know if it solves the problem that @pjreed brought up in #298?

clalancette:
> @clalancette - Everything in this MR looks fine. Do you know if it solves the problem that @pjreed brought up in #298?

I believe it should, but a confirmation would be great.

pjreed:
Yep, I just tested it out and it looks much better now.  Thanks!

clalancette:
> Yep, I just tested it out and it looks much better now. Thanks!

Fantastic, thanks for testing!

PR #310, opened 2019-11-08
Fixes pointed out by clang-tidy

clalancette:
This is a bunch of small fixes pointed out when running clang-tidy against the codebase.  Most of them are minor, though there is one fix in here that changes the calculation of the intensity to match the documentation.  It's probably easiest to review this commit-by-commit, as I've split them up logically.


clalancette:
There are a lot of changes here, but most of them are cleanup or making things more nice.  There are 2 important bugfixes that I'll call out separately:

* 7c35acd - This changes the intensity calculation to follow the guidance in the VLP datasheet.
* 098058b - This clears out the data in the pointcloud before using it, which gets rid of some odd artifacts when using organize_cloud along with view_width.

Both of these could be considered to be backported to the ROS 1 version.

clalancette:
> Everything looks good to me. Most of them are pretty straight-forward. Since this changes the intensity reading, have you tried it on a sensor? Did you notice anything obvious?

I did try it out on both the VLP16 and VLP32C I have, and nothing was obviously wrong in the RViz2 visualization.  However, I'm honestly not entirely sure how to test that it is doing the "right" thing; any thoughts there?

JWhitleyAStuff:
Now that I think about it, the only way would be using a target with a known reflectivity and I doubt you have one of those lying around. If intensity appears to still be working in general, I'm OK with merging.

clalancette:
> Now that I think about it, the only way would be using a target with a known reflectivity and I doubt you have one of those lying around. If intensity appears to still be working in general, I'm OK with merging.

Good assumption :).  Thanks for the review and merge!

Issue #292, opened 2019-10-09
Laser scan nodelet reports "PointCloud2 fields in unexpected order. Using slower generic method."


VorpalBlade:
**Please complete the following information:**
 - OS and Version: Ubuntu 16.04
 - ROS Version: Kinetic
 - Built from Source
 - Version: Commit 720829573076bf7d7edd4b1de344924fad909049

**Describe the bug**
Using a VLP-16, I get a warning from velodyne_laserscan that the fields are not in the expected order:

```
[ WARN] [1570599342.210624397]: VelodyneLaserScan: PointCloud2 fields in unexpected order. Using slower generic method.
```

This did not happen using the published version of the stack available in Kinetic.

Looking into the point cloud message it's structure is:

```yaml
fields: 
  - 
    name: "x"
    offset: 0
    datatype: 7
    count: 1
  - 
    name: "y"
    offset: 4
    datatype: 7
    count: 1
  - 
    name: "z"
    offset: 8
    datatype: 7
    count: 1
  - 
    name: "intensity"
    offset: 12
    datatype: 7
    count: 1
  - 
    name: "ring"
    offset: 16
    datatype: 4
    count: 1
```

The laser scan nodelet however expects the intensity to be at index 16 and ring at index 20. It doesn't check for anything at index 12.

**To Reproduce**
1. Connect VLP-16
2. `roslaunch velodyne_driver nodelet_manager.launch model:=VPL16 rpm:=1200`
3. `roslaunch velodyne_pointcloud cloud_nodelet.launch calibration:=<your calibration file> min_range:=0.1`
4. `roslaunch --screen velodyne_pointcloud laserscan_nodelet.launch`
5. See warning

**Expected behavior**
The laser scan nodelet should be able to use the fast method.

**Additional context**
VLP-16 firmware version is 3.0.41.1, which was the latest as of writing this.

JWhitleyAStuff:
This should be fixed in #316. Please test with the latest `master` branch and see if the issue is fixed for you. If not, please re-open this ticket.

PR #316, opened 2019-11-21
Add laserscan support for new PointXYZIR structure

kmhallen:
Untested, but it's just a simple change.

Addresses issue #292


JWhitleyAStuff:
@kmhallen - There's a good amount of code duplication going on here. Can you refactor to store a `ring_idx` and `intensity_idx` and then just fill them into a single copy of the `if` statement?

PR #318, opened 2019-11-25
Getting model param for timings from private node handle.

JWhitleyAStuff:
The current implementation in `rawdata.cc` gets the `model` parameter from a public node handle which is referencing a static node name. Since the node name is dynamic and can be namespaced, this rarely works. Changed to use the passed-in private node handle to get the parameter dynamically.

### DIFF
--- a/velodyne_pointcloud/src/lib/rawdata.cc
+++ b/velodyne_pointcloud/src/lib/rawdata.cc
@@ -196,9 +196,7 @@ inline float SQR(float val) { return val*val; }
   /** Set up for on-line operation. */
   boost::optional<velodyne_pointcloud::Calibration> RawData::setup(ros::NodeHandle private_nh)
   {
-    // ros::NodeHandle nh_driver("private_nh.getNamespace().substr(0, private_nh.getNamespace().length()-6)+"_driver"");
-    ros::NodeHandle nh_driver("/velodyne_nodelet_manager_driver");
-    nh_driver.param("model", config_.model, std::string("64E"));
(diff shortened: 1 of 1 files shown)


VorpalBlade:
This does appear to work, with a caveat that might be a breaking change in functionality.

At first I got these lines when starting the node:
```
[ WARN] [1574755724.885047485]: Timings not supported for model 
[ WARN] [1574755724.885070989]: NO TIMING OFFSETS CALCULATED. ARE YOU USING A SUPPORTED VELODYNE SENSOR?
```
This is the relevant fragment of the launch file I had:
```
  <group ns="sensors/laser">
    <include file="$(find velodyne_driver)/launch/nodelet_manager.launch">
      <arg name="model" value="VLP16" />
      <arg name="rpm" value="1200" />
    </include>
    <include file="$(find velodyne_pointcloud)/launch/cloud_nodelet.launch">
      <arg name="calibration" value="$(find oru_ipw_description)/calibration/vlp16_calib.yaml" />
      <arg name="min_range" value="0.1" />
    </include>
  </group>
```

Adding `<arg name="model" value="VLP16" />` to the cloud_nodelet.launch fixed the issue.

However, the model parameter is not documented at https://wiki.ros.org/velodyne_pointcloud

JWhitleyAStuff:
@VorpalBlade - Thanks for the information. That parameter was only recently added to the Cloud Nodelet, specifically for this feature. I'll get it added to the documentation.

JWhitleyAStuff:
@VorpalBlade - Documentation on the Wiki has been updated with the `model` parameter. Thanks!

PR #322, opened 2019-12-06
Fix ring retrieval from float pointcloud iterator.

hidmic:
Precisely what the title says. I stumbled upon this while trying to use the `velodyne_laserscan` node along with `velodyne` Gazebo plugins. `r` would always be `0`. I *believe* (do correct me if I'm wrong) that this is what the original author had in mind. 

Tagging @clalancette for awareness.

### DIFF
--- a/velodyne_laserscan/src/velodyne_laserscan.cpp
+++ b/velodyne_laserscan/src/velodyne_laserscan.cpp
@@ -179,9 +179,7 @@ void VelodyneLaserScan::recvCallback(const sensor_msgs::msg::PointCloud2::Shared
       scan->intensities.resize(kSize);
 
       for (sensor_msgs::PointCloud2ConstIterator<float> it(*msg, "x"); it != it.end(); ++it) {
-        const uint16_t r =
-          (static_cast<const uint16_t>(it[5]) << 8) |
-          static_cast<const uint16_t>(it[6]);
+        const uint16_t r = *(reinterpret_cast<const uint16_t *>(&it[5]));
 
         if (r == ring) {
           const float x = it[0];  // x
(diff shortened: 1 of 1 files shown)


velodyne_laserscan/src/velodyne_laserscan.cpp
```diff
@@ -179,9 +179,7 @@ void VelodyneLaserScan::recvCallback(const sensor_msgs::msg::PointCloud2::Shared
       scan->intensities.resize(kSize);
 
       for (sensor_msgs::PointCloud2ConstIterator<float> it(*msg, "x"); it != it.end(); ++it) {
-        const uint16_t r =
-          (static_cast<const uint16_t>(it[5]) << 8) |
-          static_cast<const uint16_t>(it[6]);
+        const uint16_t r = *(reinterpret_cast<const uint16_t *>(&it[5]));
```
clalancette: So, if I'm reading this correctly, `it[5]` returns a float value (reference, I guess, but that doesn't make much difference here).  We then take the address of that, so we now have a `float *`.  Then we `reinterpret_cast` that to a `uint16_t *`, and then we dereference that and store it into the `uint16_t`.  But isn't that meaningless?  That is, casting a `float *` to a `uint16_t *` just gives you some of the bits of the mantissa (or however the float is laid out bitwise, I forget the details).  Am I missing something here?

hidmic: Your reasoning is correct. The thing is, `ring` actually is a `uint16_t` and it is layed out in memory as such. Reading it as a `float` is what is meaningless, for the reasons you've described. I presume it was written this way to spare another iterator (a performance boost?) and I didn't want to change that.

clalancette: Oh!  Wow, that is subtle.  Thanks for the explanation, that is helpful.  Would you mind adding a comment above this code basically explaining that?  I think it will save confusion for someone in the future.  Otherwise, looks good to me.

hidmic: Done in https://github.com/ros-drivers/velodyne/pull/322/commits/12f6263e9ec279ccb8b4a2e48bddc2b63774068a.

hidmic:
@JWhitleyAStuff friendly ping.

JWhitleyAStuff:
Sorry, been travelling. LGTM too.

clalancette:
> Sorry, been travelling. LGTM too.

Thanks for the review.  I'll merge now.

PR #331, opened 2020-02-18
Fix dependencies in package.xml

clalancette:
Ensure that we depend on ament_cmake_ros as appropriate.

### DIFF
--- a/velodyne_driver/package.xml
+++ b/velodyne_driver/package.xml
@@ -20,7 +20,7 @@
   <author>Yaxin Liu</author>
   <author>Andreas Klintberg</author>
 
-  <buildtool_depend>ament_cmake</buildtool_depend>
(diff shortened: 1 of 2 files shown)


JWhitleyWork:
@clalancette - I'm actually not familiar. What's the difference between `ament_cmake` and `ament_cmake_ros`?

clalancette:
> @clalancette - I'm actually not familiar. What's the difference between `ament_cmake` and `ament_cmake_ros`?

Essentially, `ament_cmake_ros` adds a few defaults to `ament_cmake` that are more consistent for ROS nodes.  Things like building shared libraries by default, etc.  In the end, it is not super important whether we use `ament_cmake` or `ament_cmake_ros`, but the package.xml and CMakeLists.txt have to agree (otherwise you can get build failures on the buildfarm).

PR #333, opened 2020-02-21
More fixing of dependencies.

clalancette:
1.  Ensure that package.xml lists all of the test dependencies
needed.
2.  Ensure that velodyne_cloud_types depends on rclcpp.
3.  Make sure test_row_step depends on geometry_msgs and
sensor_msgs.  It doesn't have any *direct* dependencies on
them, but it includes datacontainerbase.hpp which does have
direct dependencies.

### DIFF
--- a/velodyne_pointcloud/package.xml
+++ b/velodyne_pointcloud/package.xml
@@ -32,7 +32,6 @@
   <depend>yaml-cpp</depend>
 
   <test_depend>ament_cmake_gtest</test_depend>
-  <test_depend>ament_cmake_xmllint</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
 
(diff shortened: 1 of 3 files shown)


velodyne_pointcloud/package.xml
```diff
@@ -31,6 +31,8 @@
   <depend>velodyne_msgs</depend>
   <depend>yaml-cpp</depend>
 
+  <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>ament_cmake_xmllint</test_depend>
```
JWhitleyWork: This is included in `ament_lint_common`. See https://github.com/ament/ament_lint/blob/608c41e76023c22c82c014ed0e4c755cb8b05b6a/ament_lint_common/package.xml#L24

clalancette: Fixed in c33ff77, thanks.

Issue #287, opened 2019-09-02
Question: PointXYZIR definition was deleted, why?


beetleskin:
The point definition was valid for all Velodyne Pointclouds. Workin with pcl requires to define such a point, so why removing it? How I have to guess the point type in order to work with the driver's pointcloud and pcl.

I think the definition was removed here: https://github.com/ros-drivers/velodyne/commit/b4ad0ae9b601d1f9fa234a1bbaaa8cbc179e4bcb.

----

See http://wiki.ros.org/velodyne_pointcloud?distro=melodic#C.2B-.2B-_API.

JWhitleyAStuff:
Removing the PCL dependency significantly increased the speed of the node. In addition, it allowed us to break away from the PCL definitions of what point clouds should contain and allows for adding more containers for point clouds (such as the OrganizedCloudXYZIR). The result is still published as a standard ROS [sensor_msgs/PointCloud2](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/PointCloud2.html) which contains enough information for the cloud to be converted to a PCL cloud using [fromROSMsg](http://docs.ros.org/hydro/api/pcl_conversions/html/namespacepcl.html#af662c7d46db4cf6f7cfdc2aaf4439760). Where are you having a problem with the container? What are you trying to do with it and from where?

beetleskin:
As stated, fromROSMsg creates a `PointCloud<T>`. One needs the T. If you get the T wrong (wrong packing, order, atomic types) the conversion does not work without further notice (the cloud is just scrambled) afaik. 
So, with removing the PointXYZIR, every user of this driver working with PCL will have to implement PointXYZIR on his own. I found it very convenient and straight-forward to use. Also, there is simply no other choice for the point type other than PointXYZIR, right? So why not providing it.

Now users will be puzzled on how to convert the  message correctly. As for me, I just copied the old definition from the driver over to my code.

beetleskin:
Despite that, a lot of users build against this type right now (provided by the package released to kintetic/melodic atm). Removing it is a big ABI breaking change. Everyone using it needs to adapt. I'm not saying one should not do that, but most ros packages avoid that and I wonder what the reason was.

Or is there another way?

JWhitleyAStuff:
@spuetz - It looks like you were the implementer in #214. Do you have any feedback on this?

spuetz:
[...]

PR #335, opened 2020-03-24
Velodyne pcl

spuetz:
Regrading #287 The `velodyne_pcl` ROS package restores the conversion file `point_types.h`.
The README explains how to use it.

The following has been done:
- Fixed assignment of the time value in the organized point cloud container
- Added `velodyne_pcl` ROS package with the `point_types.h` header file.
- Added a RAEDME.md file to explain how to convert velodyne ROS point clouds to PCL
- Renamed both containers to cover the previously added time property.

### DIFF
--- a/velodyne_pcl/CMakeLists.txt
+++ b/velodyne_pcl/CMakeLists.txt
@@ -1,7 +1,13 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(velodyne_pcl)
 
-find_package(catkin REQUIRED COMPONENTS pcl_ros)
+find_package(catkin REQUIRED COMPONENTS pcl_ros roslint)
+
+roslint_cpp()
+
+if(CATKIN_ENABLE_TESTING)
+  roslint_add_test()
+endif()
 
 catkin_package(
   INCLUDE_DIRS include
--- a/velodyne_pcl/CMakeLists.txt
+++ b/velodyne_pcl/CMakeLists.txt
@@ -1,0 +1,20 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(velodyne_pcl)
+
+find_package(catkin REQUIRED COMPONENTS pcl_ros)
+
+catkin_package(
+  INCLUDE_DIRS include
+  CATKIN_DEPENDS pcl_ros
+)
+
+include_directories(
+  include
+  ${catkin_INCLUDE_DIRS}
+)
+
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.h"
+)
+
--- a/velodyne_pcl/CMakeLists.txt
+++ b/velodyne_pcl/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.5)
 project(velodyne_pcl)
 
 find_package(catkin REQUIRED COMPONENTS pcl_ros roslint)
--- a/velodyne_pcl/package.xml
+++ b/velodyne_pcl/package.xml
@@ -13,7 +13,8 @@
   <url type="repository">https://github.com/ros-drivers/velodyne</url>
   <url type="bugtracker">https://github.com/ros-drivers/velodyne/issues</url>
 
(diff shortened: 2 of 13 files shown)


spuetz:
@JWhitleyWork Could you test the conversion? Cheers.

JWhitleyWork:
@spuetz Just two requests if you could: 

1. Please add a nodelet which simply does the conversion and a launch file that runs the nodelet with parameters for a nodelet manager and input and output topics. 
2. Please add `roslint` to the package and the associated test.

JWhitleyWork:
Thanks again for working on this.

spuetz:
The conversion is also automatically done by the subscriber and the publisher. See http://wiki.ros.org/pcl/Overview#Publishing_and_subscribing_to_point_cloud_messages
I think it would make sense to just provide a test where we know which data is in the could and then compare it after conversion.
Would be good if someone could write the test, I don't have much time... and I think you are more into the testing of your laser scan data with the pcap data.

JWhitleyWork:
Thanks for the clarification, @spuetz.

JWhitleyWork:
@spuetz I'll add `roslint` to the package.  However, I've also just realized that this MR breaks API in another way by doing away with the `PointCloudXYZIR` type. While I agree that the type name should match the contents, this name change breaks existing use of the type by [many, many](https://github.com/search?q=velodyne_pointcloud+PointXYZIR&type=Code) examples on Github. This change can be introduced in Noetic, but not in the Melodic LTS. Please revert this commit.

spuetz:
Someone introduced the time field in that container, instead of creating a new one. That mainly caused the problem. Cleanest solution would be to add another container without the time field and to add the corresponding point type here.

These depending packages, you mentioned, they also need to change the include file to `velodyne_pcl/point_types.h`. Additionally they need to install the `velodyne_pcl` package.

spuetz:
@JWhitleyWork  Any news here? See my comment above.

JWhitleyWork:
@spuetz Sorry I haven't responded lately. Yes, I understand that users will need to change the include path and depend on `velodyne_pcl`. Since these changes are necessary already, changing the point type doesn't seem like that big of a deal.

I'm creating a PR to your branch right now to add `roslint`.

JWhitleyWork:
PR Created for `roslint`: https://github.com/spuetz/velodyne/pull/1

YoshuaNava:
@JWhitleyWork @spuetz I'm looking forward to this PR being merged and would be glad to help test if needed. I have a VLP-16 that I can use for tests.

JWhitleyWork:
@YoshuaNava I would really appreciate that! Could you possibly test this prior to merge? If you clone from @spuetz fork of the repository and check out his `velodyne_pcl` branch, you should be able to build and test it. Please let me know how it goes and I'll merge it immediately after!

YoshuaNava:
@JWhitleyWork Can you send me a short list of the functionality I should text, and what is expected from each test?

velodyne_pcl/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+cmake_minimum_required(VERSION 2.8.3)
```
JWhitleyWork, line 1: Since this is now targeting Noetic, please update to 3.5.

velodyne_pcl/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<package format="2">
```
JWhitleyWork, line 2: Please update to Format 3 and include the XML namespace declaration. e.g.

```
<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
```

velodyne_pcl/package.xml
```diff
@@ -0,0 +1,20 @@
...
+  <name>velodyne_pcl</name>
+  <version>0.0.0</version>
```
JWhitleyWork, line 4: Please set to version 1.5.2 to match existing version for ease of upgrade with `catkin` tools.

velodyne_pcl/package.xml
JWhitleyWork: Thanks for adding the URL. However, you need to change the `<package>` tag to match (format="3").

JWhitleyWork:
Last item: can you rebase on master so it uses Noetic CI?

Issue #342, opened 2020-05-20
Confusing naming of TF frame parameters in velodyne_pointcloud package


AndreasR30:
You use the following parameter names to specify TFs:
1. **frame_id**: "fixed frame of reference for point clouds" (According to TransformNode.cfg)
  1.1. Is stored in the code to _target_frame_ in reconfigure callback:
`config_.target_frame = tf::resolve(tf_prefix_, config.frame_id);`
  1.2. So why it has a different name in dynamic reconfigure?
2. **fixed_frame**: "The desired input frame" (According to CloudNode.cfg)
  2.1. Does not seem to be used anywhere
  2.2. Input frame should be clear from the header of the _velodyne_packets_ message
  2.3. So there should be no need to specify the input frame explicitly, right?
3. **target_frame**: "The desired output frame" (According to CloudNode.cfg)
  3.1. This is the frame used for inertial correction of the velodyne packets in Transform Node:
`tf_ptr->lookupTransform(config_.target_frame, cloud.header.frame_id, time, transform);`
  3.2. So for the Transform Node it is typically set to something like "odom"
  3.2. For both, Transform and Cloud Node: This is also the output frame in the header of the final _velodyne_points_ message

I would propose the following naming convention:
1. **frame_id**: **Just** the input frame for the **velodyne_driver**
  1.1. This frame is stored into the header of the _velodyne_packets_ message
2. **fixed_frame**: This should be the reference frame used for inertial correction
  2.1. As it is **fixed** in the world frame (required to perform inertial correction)
  2.2. Consequently set to something like "odom"
3. **target_frame**: This is the target/ouput frame of the final _velodyne_points_ message
  3.1. Nice feature to specify a frame to which the pointcloud (PC) should be finally transformed
  3.2. For example to transform the PC back to velodyne frame after performing inertial correction (currently the oulput frame would be forcibly something like "odom")

AndreasR30:
Maybe even better (more intuitive) names for the TFs would be (given my proposals):
- **frame_id** -> **sensor_frame**
- **fixed_frame** -> **odom_frame** (only required for Transform Node)
- **target_frame** -> **target_frame** or ouput_frame

Maybe another reason to use the new names would be to still support the old behavior (for backwards compatibility) while at the same time supporting the proposed behavior with the new names

JWhitleyWork:
@AndreasR30 I very much agree with this but don't have time to create a Pull Request to apply the change. Could you please create one and I will review?

[...]

PR #344, opened 2020-05-26
Unify tf frame parameters between transform and cloud nodes

AndreasR30:

Fixes ros-drivers/velodyne#342


cynosure4sure:
This is really useful inclusion. Can it be merged soon?

AndreasR30:
Unfortunately Github's file size limit is limited to 10MB so I published a very short explanatory video at YouTube: https://youtu.be/Ji9btFjtnJM

cynosure4sure:
> Unfortunately Github's file size limit is limited to 10MB so I published a very short explanatory video at YouTube: https://youtu.be/Ji9btFjtnJM

Thank you for sharing. Two questions:
1. It is difficult to see the affect of motion compensation in the video. Have you compared the compensated and uncompensated clouds to verify they are being transformed correclty?
2. What is the frequency at which your 'odom' frame is being published? As the car is moving I assume it would be at a very high update rate.

AndreasR30:
>     1. It is difficult to see the affect of motion compensation in the video. Have you compared the compensated and uncompensated clouds to verify they are being transformed correclty?

You can see in the video, that as soon the field for `fixed_frame` is empty the rings of the point clouds transform from a helix-like shape to a regular circle (gaps in the back of the vehicle disappear). At this moment ego motion compensation is turned **off**. Compared to the last point of a ring, the first point is shifted backwards by ego motion compensation because the vehicle has moved forwards during a single scan.

>     2. What is the frequency at which your 'odom' frame is being published? As the car is moving I assume it would be at a very high update rate.

Currently our odom frame is published at 100Hz. Our velodyne sensor rotates with 10 rotations per second (600rpm). So there are 10 TF "updates" for a full rotation or a "update" every 36deg of rotation. This is sufficient for our use case as we are not driving that fast. It would be no problem to publish the odom frame at a higher rate for highly dynamic scenarios (especially with high turn rates of the ego vehicle). Currently the transformations are only done per packet (contains 384 points == 6 shots of the 64 diodes for our Velodyne 64E_S2). Therefore maximum meaningful rate to publish odom frame would be approx. 3470Hz for our  64E_S2 because it publishes ~347 packets per rotation and our sensor rotates at 10Hz.

cynosure4sure:
Thank you for the very informative response and thank you for your work. I hope it gets merged soon.

UniBwTAS:
We also think that this pull request is very useful as it reduces duplicate code and, more importantly, fixes some bugs with the frame_id names.

UniBwTAS:
@JWhitleyWork Do you have any plans to review this pull request?

JWhitleyWork:
>@JWhitleyWork Do you have any plans to review this pull request?

I apologize for the long delay. My job has been overwhelming lately. I will review it today.

To what extent has this been tested and on which sensors?

cynosure4sure:
@JWhitleyWork I have tested this pull request on VLP-16. @AndreasR30 has tested this on Velodyne 64E_S2 as shown in their shared video above.

UniBwTAS:
To avoid cunfusion: @AndreasR30 works at our institute and will communicate just through this account from now on.

velodyne_pointcloud/launch/32e_points.launch
```diff
@@ -40,13 +40,11 @@
...
 
-  <!-- start cloud nodelet -->
-  <include file="$(find velodyne_pointcloud)/launch/cloud_nodelet.launch">
+  <!-- start transform nodelet -->
+  <include file="$(find velodyne_pointcloud)/launch/transform_nodelet.launch">
     <arg name="model" value="32E"/>
     <arg name="calibration" value="$(arg calibration)"/>
     <arg name="manager" value="$(arg manager)" />
-    <arg name="fixed_frame" value="$(arg frame_id)" />
-    <arg name="target_frame" value="$(arg frame_id)" />
```
JWhitleyWork, line 49: Why would you want to remove these instead of exposing them at the top level?

UniBwTAS: My intention was that these parameters are not necessary to get the old CloudNode(let) behavior, because fixed_frame and target_frame were not really used in convert.cc, but just in transform.cc. So the old CloudNode behavior was to publish a non-inertial corrected point cloud in the sensor frame (obtained from the VelodyneScan.msg message). 
In the new implementation this is the default behaviour. So it is not necessary to pass these arguments.
But you're right, maybe we should put them back to show the user, that these options/params do exist. However, this time with empty strings, i.e. `<arg name="target_frame" value="" />`

JWhitleyWork: Yes, agreed. Please add them back with empty strings.

JWhitleyWork:
@AndreasR30 Since `master` now targets Noetic and has new CI to match, please rebase your branch on `master` so CI will build in Noetic.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
```diff
@@ -157,12 +148,14 @@ class DataContainerBase
     eigen_vec(2) = tf_vec[2];
   }
 
-  inline bool computeTransformation(const ros::Time& time)
+  inline bool lookupHTM(Eigen::Affine3f& htm, const std::string& target_frame,
```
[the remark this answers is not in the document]
UniBwTAS: **H**omogeneous **T**ransformation **M**atrix, i.e. transform matrix for homogeneous coordinates. It combines rotation and translation.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
```diff
@@ -181,13 +174,39 @@ class DataContainerBase
...
-    Eigen::Vector3f p = transformation * Eigen::Vector3f(x, y, z);
+    Eigen::Vector3f p = Eigen::Vector3f(x, y, z);
+    if (!config_.fixed_frame.empty())
+    {
+      p = htm_to_fixed * p;
+    }
+    if (!config_.target_frame.empty())
+    {
+      p = htm_to_target * p;
+    }
```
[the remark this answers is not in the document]
UniBwTAS: Only perform a transform when fixed_frame is specified. It should be just specified if one wants to use ego motion compensation. Otherwise it should be left empty in the params. Note that `htm_to_fixed` was only calculated when fixed_frame is not empty (see `computeTransformToFixed`)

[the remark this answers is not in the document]
UniBwTAS: Same holds for target frame

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
```diff
@@ -140,12 +137,6 @@ class DataContainerBase
     config_.min_range = min_range;
     config_.fixed_frame = fixed_frame;
     config_.target_frame = target_frame;
-
-    config_.transform = fixed_frame.compare(target_frame) != 0;
-    if (config_.transform && !tf_ptr)
-    {
-      tf_ptr = boost::shared_ptr<tf::TransformListener>(new tf::TransformListener);
-    }
```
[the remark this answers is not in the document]
UniBwTAS: I'm not sure whether really a relevant amount of Memory and CPU is wasted by using a TF Listener. Compared to processing millions of points I would say this is negligible. But I can put it back when you think this is very important.

velodyne_pointcloud/include/velodyne_pointcloud/transform.h
```diff
@@ -76,11 +76,9 @@ class Transform
   boost::shared_ptr<dynamic_reconfigure::Server<velodyne_pointcloud::TransformNodeConfig>> srv_;
   void reconfigure_callback(velodyne_pointcloud::TransformNodeConfig& config, uint32_t level);
 
-  const std::string tf_prefix_;
   boost::shared_ptr<velodyne_rawdata::RawData> data_;
-  message_filters::Subscriber<velodyne_msgs::VelodyneScan> velodyne_scan_;
-  ros::Publisher output_;
-  boost::shared_ptr<tf::MessageFilter<velodyne_msgs::VelodyneScan>> tf_filter_ptr_;
```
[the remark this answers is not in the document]
UniBwTAS: I renamed it because I think it is easier to read. So the node subscribes to a Velodyne**Scan**.msg and publishes a Point**Cloud**2.msg.

velodyne_pointcloud/src/conversions/transform.cc
```diff
@@ -43,38 +42,15 @@ namespace velodyne_pointcloud
...
-                                  config_.num_lasers, data_->scansPerPacket(), tf_ptr_));
-    }
-    else
-    {
-      container_ptr = boost::shared_ptr<PointcloudXYZIR>(
-          new PointcloudXYZIR(config_.max_range, config_.min_range,
-                              config_.target_frame, config_.fixed_frame,
-                              data_->scansPerPacket(), tf_ptr_));
-    }
-
```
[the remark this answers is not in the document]
UniBwTAS: It is not necessary (dublicate code) because the reconfigure callback is always executed with the params defined in the .launch file as soon a node starts.

velodyne_pointcloud/src/conversions/transform.cc
```diff
@@ -134,22 +112,35 @@ namespace velodyne_pointcloud
...
     {
-      container_ptr->computeTransformation(scanMsg->packets[i].stamp);
-      data_->unpack(scanMsg->packets[i], *container_ptr,  scanMsg->header.stamp);
+      // calculate individual transform for each packet to account for ego
+      // during one rotation of the velodyne sensor
+      if(!container_ptr->computeTransformToFixed(scanMsg->packets[i].stamp))
+      {
+        // fixed frame not available
+        return;
+      }
```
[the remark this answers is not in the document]
UniBwTAS: `computeTransformToFixed` also returns `true` when the `fixed_frame` is left empty in the params.

UniBwTAS:
> Further, we should go for tf2 here
I agree.

> I also would prevent to check if computeTransformToFixed returns true all the time.
I think it is necessary to check if the transform is available. Otherwise the program will not work properly.

> Additionally, we should watch out for the transform with respect to the time stamp related to each single point, and not only packets wise.
I'm really not sure whether it is a good idea to call lookupTransform and to calculate a transform matrix millions of times per second. This will introduce a very, very large CPU overhead for minimal win. Furthermore one would have to publish the odometry frame (or something similar) millions of times per seconds, which introduces additional overhead. We think this error introduced by a per-packet-calculation is negligible to other errors.

JWhitleyWork:
@UniBwTAS Please rebase on the current `master` which now targets `noetic` or re-target this PR to `melodic-devel` which targets Melodic and Kinetic.

velodyne_pointcloud/include/velodyne_pointcloud/transform.h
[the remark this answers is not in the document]
JWhitleyWork: I agree with @spuetz. Please separate new functionality (the merging of the two files) from "code clean-up" (renaming variables).

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
```diff
@@ -181,13 +174,39 @@ class DataContainerBase
...
+  }
+
+  inline bool computeTransformToFixed(const ros::Time &packet_time)
+  {
+    std::string &source_frame = cloud.header.frame_id;
+    return !(!config_.fixed_frame.empty() &&
+             !lookupHTM(htm_to_fixed, config_.fixed_frame, source_frame,
+                        packet_time));
+  }
+
```
spuetz, line 198: Can you explain why we need two transformations here. The `fixed_frame` should be defined by the user as the static coordinate frame, i.e. the root of the tf tree. The `target_frame` is the goal coordinate frame to which we want to transform the points, this can of course also be configured to be the `fixed_frame`. The `source_frame` should be the `sensor_frame` which is for example defined in the URDF. Another thing are the two options 1.) transform packet wise or 2.) transform once per cloud. Am I right? Please explain a bit here, if I am wrong. What am I missing here?

spuetz: I mean you are separating the transformation into two parts `fixed_frame`->`target_frame`, or `cloud.header.frame_id`->`target_frame`, where the `cloud.header.frame_id` is set either to the `target_frame` in line 121 or to `fixed_frame` in line 125 and the second transformation `source_frame` -> `fixed_frame'  where the `source_frame` is again the `cloud.header.frame_id`, so line 121 or 125. Could you maybe explain this and make some comments in the code or a simple sketch. Thank you.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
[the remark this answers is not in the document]
UniBwTAS: Ok I can rename it.

velodyne_pointcloud/include/velodyne_pointcloud/transform.h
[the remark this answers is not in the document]
UniBwTAS: Ok I will undo these renamings.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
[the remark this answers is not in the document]
UniBwTAS: Ok I can restore this behavior.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
UniBwTAS: Maybe it helps to read through issue #342  (including watching the video) if you haven't already.

The idea is that the point cloud has to be transformed packet-wise (actually even better point-wise based on its timestamp, however this would be extremely CPU intensive) into the odom frame based on the packet timestamp. At this point one automatically obtains a ego motion compensated point cloud where the point coordinates are specified w.r.t. odom. This is what `computeTransformToFixed` is intended for. When fixed_frame is not empty it computes `htm_to_fixed` for subsequent use. This matrix transforms points from sensor coords (already stored in `cloud.header.frame_id` beforehand, compare line 104) to the fixed frame.

Points in odom coords are sometimes undesired, especially when someone wants to perform downstream segmentation tasks. In this case one typically wants the points to be back in sensor coordinates, however this time ego motion compensated. Transforming the whole point cloud given a single timestamp (single transformation matrix for whole point cloud) achieves this. This point cloud is different to directly converting to target frame:

notes

This is what `computeTransformToTarget` is intended for. Here the source frame is here selected based on the fact whether ego motion compensation was performed (odom) or not (points are still in sensor frame, stored inside `cloud.header.frame_id` in line 104). Lines 121 and 125 are called after all these transformation when the point cloud is finished, so they have no effect here. You are right when you say that it is a little bit confusing that the string of the sensor frame is stored/passed in the point cloud message and not explicitly passed in a dedicated parameter/variable. However, this is a thing that was already implemented this way before my pull request and you already made clear that I should not perform any refactoring in this pull request.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
```diff
@@ -141,53 +146,86 @@ class DataContainerBase
     config_.fixed_frame = fixed_frame;
     config_.target_frame = target_frame;
 
-    config_.transform = fixed_frame.compare(target_frame) != 0;
-    if (config_.transform && !tf_ptr)
+    // only use somewhat resource intensive tf listener when it is necessary
+    if (!fixed_frame.empty() || !target_frame.empty())
```
JWhitleyWork, line 150: Shouldn't this be `if (!fixed_frame.empty() || (!target_frame.empty() && target_frame != sensor_frame))`? Seems silly to create the buffer and listener if the sensor frame and the target frame are the same and I don't think it's clear anywhere that the target frame should be left empty if you just want the sensor frame to be passed through unaltered.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
```diff
@@ -141,53 +146,86 @@ class DataContainerBase
...
-  }
-
-  inline bool computeTransformation(const ros::Time& time)
-  {
-    tf::StampedTransform transform;
+    geometry_msgs::TransformStamped msg;
     try
     {
-      tf_ptr->lookupTransform(config_.target_frame, cloud.header.frame_id, time, transform);
+      msg = tf_buffer->lookupTransform(target_frame, source_frame, time, ros::Duration(0.2));
```
JWhitleyWork, line 173: It's possible that this function could be called when no TF buffer has been created. Should be checked for existence with the shared pointer's bool semantics first.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
```diff
@@ -127,6 +124,14 @@ class DataContainerBase
...
       cloud.header.frame_id = config_.target_frame;
     }
+    else if (!config_.fixed_frame.empty())
+    {
+      cloud.header.frame_id = config_.fixed_frame;
+    }
+    else
+    {
+      cloud.header.frame_id = sensor_frame;
+    }
```
JWhitleyWork, line 134: It should be made clear somewhere (wiki.ros.org?) that fixed_frame > target_frame > sensor_frame.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
```diff
@@ -200,8 +238,11 @@ class DataContainerBase
 
 protected:
   Config config_;
-  boost::shared_ptr<tf::TransformListener> tf_ptr;  ///< transform listener
-  Eigen::Affine3f transformation;
+  boost::shared_ptr<tf2_ros::TransformListener> tf_listener;
+  boost::shared_ptr<tf2_ros::Buffer> tf_buffer;
```
JWhitleyWork, line 242: Is there a reason to use `boost` instead of `std` here?

AndreasR30: It was boost before so I used boost again. You already made clear that I should avoid refactoring within this pull request...

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
AndreasR30: You are right I will add this. I had this already in my mind.

UniBwTAS: You are right I will add this. I had this already in my mind.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
UniBwTAS: Normally there this shouldn't be possible because this function is just called when fixed_frame or target_frame are specified. However I can add an additional check to make the code more robust.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
UniBwTAS: Yes that makes sense to document this properly. I would describe it in a similar way as I did in the bullet points of [this comment](https://github.com/ros-drivers/velodyne/issues/342#issuecomment-633029124) when I described the behavior in my new version.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
UniBwTAS: Apart from this, it would be possible to use std::shared_ptr as the source is compiled with C++11 when I see it correctly.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
UniBwTAS: > fixed_frame > target_frame > sensor_frame

Not sure what you exactly mean by this. Do you mean the parameter precedence?
Then it should be:
`target_frame > fixed_frame > sensor_frame`

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
UniBwTAS: I addressed this in my latest commit.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
JWhitleyWork: Thanks. Better to write your intent in the code than leave it to chance.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
JWhitleyWork: Yes, sorry. I got the order wrong.

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
JWhitleyWork: Thanks for the clarification. I'll remove as much `boost` as I can in a future PR.

UniBwTAS:
@spuetz and @JWhitleyWork is everything ok for you now or did I miss some remarks? I would be happy when you integrate it into the master branch.

JWhitleyWork:
@UniBwTAS Thank you for working on this! Can you please create another PR to apply the same changes to the `melodic-devel` branch?

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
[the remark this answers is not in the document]
spuetz: @JWhitleyWork This hasn't been addressed?

Issue #340, opened 2020-04-29
ROS Melodic and the Broken API on Master


JWhitleyWork:
Several months ago, I merged the PR #214 into `master`. This was an excellent, very efficient PR but what I didn't realize at the time was that it broke the API from the Melodic release of the `velodyne` packages. On the plus (and minus) side, I have yet to release a new Melodic version of this repository so hell has been kept from breaking loose. Since Melodic is an LTS ROS release and there are likely a lot of people who wouldn't be very happy with a package update breaking their downstream things, the following is my new plan of action:

1. Go back to the commit before b4ad0ae introduced the breaking changes and create a melodic branch from there.
2. Do my best to cherry-pick changes that were made after that point onto the melodic branch without breaking the API.
3. Have someone test the melodic branch.
4. Release the melodic branch into the build farm as a Melodic release.
5. Consider master to be targeted at Noetic.

Not only does this remove the need for a hack like #335, it keeps from breaking the Melodic API. Noetic will then receive the new, no-PCL API (with the PointCloud type renamed to include the new fields).

I would appreciate any feedback from @spuetz, @beetleskin, and @YoshuaNava (and anyone else that wishes to chime in).

If I don't hear any dissenting opinions, I'll get started on this work sometime in the next 2 weeks and hopefully we'll have a shiny, new Melodic release soon thereafter.

spuetz:
I think adding a bit of code should fix that issue. But also marking the code maintaining backwards compatibility as deprecated.

JWhitleyWork:
@spuetz Could you explain a little more? I don't quite understand your answer.

spuetz:
@JWhitleyWork Yes of course, sry for the short answer above. I think we could add the necessary files  like the point_types.h and the corresponding container, which is maintaining the old point cloud fields. We add these to the `velodyne_pointcloud` package and we add the pcl dependency again. Then we mark these files and usage as deprecated with a link to the velodyne_pcl package. What do you think?

JWhitleyWork:
@spuetz Assuming that you mean we do the second step for Noetic only, then yes, that would also work. Could you possibly create the PR for this proposal? I'm sorry but I'm not as familiar with the changes you made to remove PCL as I should be.

spuetz:
@JWhitleyWork I'll create a PR in the next week.

JWhitleyWork:
> @JWhitleyWork I'll create a PR in the next week.

Thank you! Sorry I haven't been able to give time to this lately.

[...]

PR #347, opened 2020-06-08
Containers and PCL point type for backwards compatibility 

spuetz:
This is a possible solution for #340 for melodic.
- It restores the pcl dependency
- The missing `point_type.h` is added again.
- Corresponding containers without a time field are added.
- The assignment of the time value has been fixed for the organized cloud container, see `organized_cloudXYZIRT.cc`.

Things to do before merging / releasing
- [ ] Someone using the driver with the `point_type.h` should test this.
- [ ] For Noetic, we should remove pcl and add a `velodyne_pcl` package (#335)
- [ ] Furthermore, we could merge the classes in `convert.h` and `transform.h`, since it more or less identical now.

### DIFF
--- a/velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
@@ -1,0 +1,46 @@
+/* -*- mode: C++ -*-
+ *
+ *  Copyright (C) 2011, 2012 Austin Robot Technology
+ *
+ *  License: Modified BSD Software License Agreement
+ *
+ *  $Id: data_base.h 1554 2011-06-14 22:11:17Z jack.oquin $
+ */
+
+/** \file
+ *
+ *  Point Cloud Library point structures for Velodyne data.
+ *
+ *  @author Jesse Vera
+ *  @author Jack O'Quin
+ *  @author Piyush Khandelwal
+ */
+
+#ifndef __VELODYNE_POINTCLOUD_POINT_TYPES_H
+#define __VELODYNE_POINTCLOUD_POINT_TYPES_H
+
+#include <pcl/point_types.h>
+
+namespace velodyne_pointcloud
+{
+  /** Euclidean Velodyne coordinate, including intensity and ring number. */
+  struct PointXYZIR
+  {
+    PCL_ADD_POINT4D;                    // quad-word XYZ
+    float    intensity;                 ///< laser intensity reading
+    uint16_t ring;                      ///< laser ring number
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW     // ensure proper alignment
+  } EIGEN_ALIGN16;
+
+}; // namespace velodyne_pointcloud
+
+
+POINT_CLOUD_REGISTER_POINT_STRUCT(velodyne_pointcloud::PointXYZIR,
+                                  (float, x, x)
+                                  (float, y, y)
+                                  (float, z, z)
+                                  (float, intensity, intensity)
+                                  (uint16_t, ring, ring))
+
+#endif // __VELODYNE_POINTCLOUD_POINT_TYPES_H
+
--- a/velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
@@ -26,9 +26,11 @@ namespace velodyne_pointcloud
   /** Euclidean Velodyne coordinate, including intensity and ring number. */
   struct PointXYZIR
   {
-    PCL_ADD_POINT4D;                    // quad-word XYZ
-    float    intensity;                 ///< laser intensity reading
-    uint16_t ring;                      ///< laser ring number
+    float x;                         ///< x coord
+    float y;                         ///< y coord
+    float z;                         ///< z coord
+    float intensity;                 ///< laser intensity reading
+    uint16_t ring;                   ///< laser ring number
     EIGEN_MAKE_ALIGNED_OPERATOR_NEW     // ensure proper alignment
   } EIGEN_ALIGN16;
 
--- a/velodyne_pointcloud/src/conversions/convert.cc
+++ b/velodyne_pointcloud/src/conversions/convert.cc
@@ -16,7 +16,9 @@
 #include "velodyne_pointcloud/convert.h"
 
 #include <velodyne_pointcloud/pointcloudXYZIR.h>
+#include <velodyne_pointcloud/pointcloudXYZIRT.h>
 #include <velodyne_pointcloud/organized_cloudXYZIR.h>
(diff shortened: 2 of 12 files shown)


JWhitleyWork:
Item 3 in your list is handled by #344. Since this functionality is necessary to get a new release, I'll merge this first and have the author of #344 rebase on your changes.

spuetz:
We should have this also in the master for `noetic` with the option of selecting the right container.
But, of course without the `point_type.h` and the pcl dependencies.
Further, we could handle the containers as plugins, using the pluginlib. Then people could use the driver and modify their container, meaning modifying what is in the cloud and how, e.g. how it is transformed, etc.
Additionally, for noetic I would remove the config flag for `organize_cloud` since this is covered by using one of the organized containers.

velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
```diff
@@ -0,0 +1,48 @@
...
+#include <pcl/point_types.h>
+
+namespace velodyne_pointcloud
+{
+/** Euclidean Velodyne coordinate, including intensity and ring number. */
+struct PointXYZIR
+{
+  float x;                         ///< x coord
+  float y;                         ///< y coord
+  float z;                         ///< z coord
```
JWhitleyWork, line 31: Just to make absolutely sure there aren't any problems, please use the PCL macro `PCL_ADD_POINT4D;` instead of these 3 types.

velodyne_pointcloud/src/conversions/convert.cc
```diff
@@ -43,19 +45,50 @@ namespace velodyne_pointcloud
...
+        container_name = default_organized_container_.first;
+        container_ptr_ = getContainer(default_organized_container_.second);
+      }
+      else
+      {
+        container_name = default_container_.first;
+        container_ptr_ = getContainer(default_container_.second);
+      }
+      ROS_WARN_STREAM("Using default container \"" << container_name << "\"!");
+      ROS_WARN_STREAM("Please use the \"container\" param with one of the following container names: " << container_list.str());
```
JWhitleyWork, line 74: Please either do one of the following:

1. Add the `container` parameter to all launch files and set it to the un-organized default of `PointCloudXYZIR`; or
2. Set this to be an `INFO` statement instead of a warning.

Having the new default value cause a warning seems problematic and could unnecessarily alarm users. I would prefer the first option since it should also be accompanied by some documentation with the possible values in the launch file.

velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
```diff
@@ -0,0 +1,48 @@
...
+  float intensity;                 ///< laser intensity reading
+  uint16_t ring;                   ///< laser ring number
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW  // ensure proper alignment
+}
+EIGEN_ALIGN16;
+
+};  // namespace velodyne_pointcloud
```
JWhitleyWork, line 38: Shouldn't we also define the PointXYZIRT in terms of a PCL structure so people can use it in the same way?

JWhitleyWork:
@spuetz Sorry, I got around to merging #344 before this. Can you please rebase and make sure all of the tests work with your changes as well as addressing my comments above?

JWhitleyWork:
@spuetz Can you please address my concerns? After that, I'll see if I can find a package to test with.

velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
spuetz: Then we would need to add a padding (It adds four folats) to the point cloud, which increases the size of the buffer and cloud object. This would result in adding another container with a padding (XYZPIR) as default. Your choice. ;)

velodyne_pointcloud/src/conversions/convert.cc
spuetz: I also prefer the first option! Will do that.

velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
spuetz: Yes for melodic that makes sense. For noetic, I'll move this to the velodyne_pcl package.

velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
spuetz: @JWhitleyWork What would you prefer? adding a useless padding to the could within a new container?

JWhitleyWork: PointXYZIR with the quad-word XYZ was the original version. Even if it had additional padding, it's what was there before so we should use it as the default and keep it the same as the original name to avoid compatibility issues. Here are the contents of this file from the released version, 1.5.2:

```
namespace velodyne_pointcloud
{
  /** Euclidean Velodyne coordinate, including intensity and ring number. */
  struct PointXYZIR
  {
    PCL_ADD_POINT4D;                    // quad-word XYZ
    float    intensity;                 ///< laser intensity reading
    uint16_t ring;                      ///< laser ring number
    EIGEN_MAKE_ALIGNED_OPERATOR_NEW     // ensure proper alignment
  } EIGEN_ALIGN16;

}; // namespace velodyne_pointcloud


POINT_CLOUD_REGISTER_POINT_STRUCT(velodyne_pointcloud::PointXYZIR,
                                  (float, x, x)
                                  (float, y, y)
                                  (float, z, z)
                                  (float, intensity, intensity)
                                  (uint16_t, ring, ring))
```

velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
JWhitleyWork: Sounds good.

JWhitleyWork:
@spuetz ping for changes.

JWhitleyWork:
@spuetz Any update here?

spuetz:
Sorry for the late reply. I'll have time in two weeks to finish this.

JWhitleyWork:
Thanks, @spuetz!

JWhitleyWork:
@spuetz ping

velodyne_pointcloud/include/velodyne_pointcloud/point_types.h
JWhitleyWork: This item still needs to be done.

JWhitleyWork:
@spuetz Still two items remaining when you can make time.

JWhitleyWork:
@spuetz Can we get this wrapped up?

JWhitleyWork:
Rebased on current `melodic-devel`. Waiting for @spuetz to complete as I just don't have the background or time.

PR #351, opened 2020-07-09
Unify tf frame parameters between transform and cloud nodes (Backport of #344 to Melodic)

JWhitleyWork:
* Unify tf frame parameters between transform and cloud nodes

* Reformat code slightly to avoid roslint errors

* Use boost bind instead of lambdas to add support for kinetic

* Fix some search and replace errors in comments

* Add fixed_frame and target_frame params back

* Use more common ROS terminology instead of 'htm'

* Undo renaming of publisher and subscriber

* Just use tf listener when necessary

* Migrate package to tf2

* Explicitly store sensor frame in a dedicated variable to make code easier to read

* Reformat code slightly to avoid roslint errors

* Avoid unnecessary transforms when sensor_frame == target_frame

* Fix comments

Co-authored-by: anre <andreas.reich@unibw.de>


JWhitleyWork:
@UniBwTAS and @spuetz Can you please look over this backport of #344 to make sure I didn't screw anything up?

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.h
```diff
@@ -141,53 +178,79 @@ class DataContainerBase
     config_.fixed_frame = fixed_frame;
     config_.target_frame = target_frame;
 
-    config_.transform = fixed_frame.compare(target_frame) != 0;
-    if (config_.transform && !tf_ptr)
-    {
-      tf_ptr = boost::shared_ptr<tf::TransformListener>(new tf::TransformListener);
-    }
```
spuetz, line 148: Hmm this should be here if we don't like to transform the cloud. It has been overlooked here https://github.com/ros-drivers/velodyne/pull/344#discussion_r452368123 too.

JWhitleyWork: @spuetz See https://github.com/ros-drivers/velodyne/pull/351/files#diff-8617d0d04cbc690d4e7529f8a263d278R144 and the conditions there for allocating the `tf_buffer` and `tf_listener`.

UniBwTAS:
I have tested it. Everything works fine.

However, it is not fully backwards compatible because cloud node was removed. Maybe we should provide some informational error message when someone tries to use cloud node in his/her .launch file. Something like:

> cloud node(let) was removed due to redundancy --> use transform node(let) with empty fixed_frame and target_frame parameters to get the old cloud node(let) behaviour

JWhitleyWork:
@UniBwTAS That's a good point. This is definitely an API change. In order to work around this, we would probably have to add 2 files:

- `velodyne_pointcloud/launch/cloud_nodelet.launch` which launches a `velodyne_pointcloud/CloudNodelet`
- `velodyne_pointcloud/src/conversions/cloud_nodelet.cc` which registers a `velodyne_pointcloud/CloudNodelet` that only produces the error message when run

Now that you mention it, though, we might just want to pass on back-porting this feature. API-breaking changes to a relatively stable LTS are generally frowned upon. Thanks for the feedback!

PR #359, opened 2020-08-26
Export DataContainers in a library so they can be used by other packages

bst:
I propose to combine all DataContainers in a library (data_containers) and make this library accessible to other catkin packages.

This allows other packages to convert velodyne_msgs to point clouds using a function call without requiring a separate transform_node or transform_nodelet.

Imagine a scenario where you recorded bag files containing raw velodyne_msgs and you want to process these bag files directly within code in a custom package in a deterministic offline mode, i.e. using the rosbag API instead of the non-determinism that comes with rosbag play and running multiple nodes. With the proposed changes, you can simply depend on velodyne_pointcloud and directly use a DataContainer of type PointcloudXYZIRT or OrganizedCloudXYZIRT to get point clouds.


bst:
The test transform_nodelet_hz_test failed, which I unfortunately cannot reproduce locally. Could this be due to non-determinism in the test setup?

JWhitleyWork:
> The test transform_nodelet_hz_test failed, which I unfortunately cannot reproduce locally. Could this be due to non-determinism in the test setup?

It's fine. I'm retrying the CI job. It should pass.

PR #360, opened 2020-09-01
Add invalid points in organized cloud

bst:
This PR fixes a bug in the current implementation when exporting organized clouds with scanners other than the VLP16.

Currently, data.addPoint() will not be called for invalid points, even though it should for organized point clouds.

In consequence, organized point clouds currently do not include invalid points, which leads to a broken neighbourhood structure. They instead include points that are either not initialized or contain information from previous scans.


JWhitleyWork:
@spuetz Would you mind giving this a look?

spuetz:
lgtm

JWhitleyWork:
@bst Could you please back-port this to the melodic branch as well?

Issue #345, opened 2020-05-26
Add .clang-format


AndreasR30:
The format of the source code is quite different to the one of my IDE. Furthermore it is not really consistent within this repository. During my last pull request, I lost a lot of time manually following the existing format. Therefore I would suggest to add a .clang-format file to the root of this repository. There is a official [ROS CppStyleGuide](http://wiki.ros.org/CppStyleGuide#Autoformatting_of_ROS_Code), which links to a .clang-format file.

JWhitleyWork:
@AndreasR30 I'm fine with updating the code to match the style guide (`roslint` is already applied but this doesn't catch all of the things that Clang does) but I think adding the file to the repo might be excessive. If you want to submit MRs to make the repo comply with https://github.com/davetcoleman/roscpp_code_format/blob/master/.clang-format, I'm happy to accept them.

YoshuaNava:
@AndreasR30 Do you think this could be addressed? +1

@JWhitleyWork Keeping the clang-format file I think is actually a good thing, as people can trigger formatting from the IDE (vscode, clion, sublime), and it also makes the repo hold a copy of its own format, in case the external repo is deleted. In any case, we could add a header to the copy of the format file saying that it is a copy of another source and it is kept there for reference. What do you think?

JWhitleyWork:
I'd be OK with that.

PR #364, opened 2020-09-14
Reformat code with clang-format file

UniBwTAS:
Add and apply "official" clang-format file from [here](https://github.com/davetcoleman/roscpp_code_format) referenced in the [ROS Wiki](http://wiki.ros.org/CppStyleGuide#Autoformatting_of_ROS_Code). Addresses #345 

Applied with following command:
`find . -iname "*.h" -o -iname "*.cpp" -o -iname "*.hpp" -o -iname "*.cc" | xargs clang-format -style=file -i`


UniBwTAS:
Unfortunately there is a small incompatibility with above clang-format file and roslint:
`/__w/velodyne/velodyne/src/velodyne_pcl/include/velodyne_pcl/point_types.h:57:  } should be on a line by itself  [whitespace/braces] [4]`
This is due to the prepocessor macro `EIGEN_ALIGN16` which resolves to `__attribute__((aligned(16)))`. clang-format puts it on the same line as the closing brace while roslint always wants the closing brace to be "a line by itself".

Not sure how to handle this issue. Make a Pull Request to roslint or to modify the clang-format file (and ideally make a pull-request to source repository of it)

YoshuaNava:
@UniBwTAS Thank you. I also proposed to add this in issue #328 

> Not sure how to handle this issue best. Make a pull request to roslint or to modify the clang-format file (and ideally make a pull request to source repository of it)?

You mean that in this case the ROS linter and clang format are doing opposite actions? This could happen again for different lines of code.

In that case, in my view, we should either:
1) keep only one linter & formatter. Even though roslint is the tool of choice for ROS, I would vote for clang format, as going in the direction of LLVM tools means that we could also have clang tidy at some point, which I think offers very powerful static analysis.
2) keep both working, but give them a more defined role, as done by http://punchthrough.github.io/styleguide/cpp.html

UniBwTAS:
> You mean that in this case the ROS linter and clang format are doing opposite actions? This could happen again for different lines of code.

Yes that is true. However, we suspect this to happen in rare cases with somewhat "exotic" language constructs like with the GCC-specific \_\_attribute\_\_ keyword above. So in our opinion it would be ok to add exceptions for such cases by disabling clang-format for a few lines.

However, we are also fine with your proposed solution to keep just clang-format and drop roslint as it the former is more versatile and more widely used.

.clang-format
```diff
@@ -0,0 +1,65 @@
...
+PenaltyBreakString: 1
+PenaltyBreakFirstLessLess: 1000
+PenaltyExcessCharacter: 1000
+PenaltyReturnTypeOnItsOwnLine: 90
+SpacesBeforeTrailingComments: 2
+Cpp11BracedListStyle: false
+Standard: Auto
+IndentWidth: 2
+TabWidth: 2
+UseTab: Never
```
YoshuaNava, line 37: +1 cool

YoshuaNava (commented):
@JWhitleyWork What is your perspective on this MR?

@UniBwTAS There is a failure in the CI. Can you check what's the issue?

UniBwTAS:
@YoshuaNava The CI just fails because of a test, more specifically the `transform_node_hz.test`. It fails because the publishing rate is too low. I had this issue already in previous commits. I think the reason is that the CPU of the CI server is sometimes too slow to process the pcap file fast enough? I don't get this test failure on my local machine. The strange thing is that the CI sometimes succeeds and sometimes not. If I would push a NoOP Commit there is a chance that the CI is successful.

JWhitleyWork:
@UniBwTAS / @YoshuaNava In general, I'm OK with the content. However, I'm not OK dropping `roslint` because there is no equivalent ROS-compatible testing that will fail if not applied correctly and can also easily be run by a user developing on thier own machine without additional knowledge (e.g. `colcon test` or `catkin run_tests` vs. a `clang-format`-specific command).

Regarding the failed test: Yes, I'm aware that it sometimes fails due to CI server hardware limitations. I'm just not comfortable dropping it lower than 3.0 Hz when it should be at 10.0 Hz. The test is essentially meaningless at lower values.

spuetz:
Reformatting I would not reformat at the price of git-line history. Also, formatting different from the chosen clange format makes sense in certain cases, in the end it should be about the readability of the code and not about adding paradigms at any price. I would advise against reformatting the whole repo. The git history is worth more per line! Reformatting per PR makes more sense.

YoshuaNava:
> I would not reformat at the price of git-line history. 
> The git history is worth more per line!

I disagree a bit here, I get the impression that the code in the repo has many different styles going on, and normally too much time is either lost or not put forward on PR to discuss code style. Automating this with clang format would make the process of implementing and reviewing contributions much easier. The codebase is also also not that big atm.

> Reformatting per PR makes more sense.

Reformatting each file per PR? That would still be nice :+1:

spuetz:
Of course you are welcome to have a different opinion on this topic. In general, I find the topic very important, but it is sometimes too dogmatic and not implemented in a goal-oriented way. Anyway, Clang can only be used on PR Diff level. In the end, all tools should help, I don't see that here, because the git-history part is more important in relation to the currently quite good code quality. In the end I recommend a Clang PR diff approach.

Translated with www.DeepL.com/Translator (free version)

spuetz:
Of course you are welcome to have a different opinion on this topic. In general, I find the topic very important, but it is sometimes too dogmatic and not implemented in a goal-oriented way. Anyway, Clang can also be used on PR Diff level. In the end, all tools should help, I don't see that here, because the git-history part is more important in relation to the currently quite good code quality. In the end I recommend a Clang PR diff approach.

YoshuaNava:
I don't have enough experience with open source. Could you explain why you would put that much emphasis in keeping the git history over having autoformatting?

spuetz:
It is easier to relate lines of code, e.g. to an issue, especially which lines together represent a functionality. I also don't say that one is more important than the other, but suggest a middle way.

PR #370, opened 2020-11-11
Add support for the velodyne Alpha Prime

UniBwTAS:
Add Support for the new Velodyne Alpha Prime with 128 lasers.

### DIFF
--- a/velodyne_driver/src/driver/driver.cc
+++ b/velodyne_driver/src/driver/driver.cc
@@ -62,7 +62,12 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
   private_nh.param("model", config_.model, std::string("64E"));
   double packet_rate;                   // packet frequency (Hz)
   std::string model_full_name;
-  if ((config_.model == "64E_S2") || 
+  if ((config_.model == "VLS128") )
+  {
+    packet_rate = 12507;     //https://github.com/velodynevls/velodyne_vls/blob/master/velodyne_driver/src/driver/driver.cc
+    model_full_name = std::string("HDL-") + config_.model;
+  }
+  else if ((config_.model == "64E_S2") ||
       (config_.model == "64E_S2.1"))    // generates 1333312 points per second
     {                                   // 1 packet holds 384 points
       packet_rate = 3472.17;            // 1333312 / 384
--- a/velodyne_driver/src/driver/driver.cc
+++ b/velodyne_driver/src/driver/driver.cc
@@ -64,8 +64,10 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
   std::string model_full_name;
   if ((config_.model == "VLS128") )
   {
-    packet_rate = 12507;     //https://github.com/velodynevls/velodyne_vls/blob/master/velodyne_driver/src/driver/driver.cc
-    model_full_name = std::string("HDL-") + config_.model;
+    packet_rate = 6253.9;     //  3 firing cycles in a data packet. 3 x 53.3 μs = 0.1599 ms is the accumulation delay per packet.
+                            //   1 packet/0.1599 ms = 6253.9 packets/second
+
+    model_full_name = config_.model;
   }
   else if ((config_.model == "64E_S2") ||
       (config_.model == "64E_S2.1"))    // generates 1333312 points per second
--- a/velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
@@ -118,6 +118,7 @@ static const uint16_t VLS128_BANK_4 = 0xbbff;
 
 static const float  VLS128_CHANNEL_TDURATION  =  2.665f;  // [µs] Channels corresponds to one laser firing
 static const float  VLS128_SEQ_TDURATION      =  53.3f;   // [µs] Sequence is a set of laser firings including recharging
+static const float  VLS128_TOH_ADJUSMENT     =  8.7f;   // [µs] μs. Top Of the Hour is aligned with the fourth firing group in a firing sequence.
 static const float  VLS128_DISTANCE_RESOLUTION=  0.004f;  // [m]
 static const float  VLS128_MODEL_ID=  161;
 
--- a/velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
@@ -109,6 +109,19 @@ static const int BLOCKS_PER_PACKET = 12;
 static const int PACKET_STATUS_SIZE = 4;
 static const int SCANS_PER_PACKET = (SCANS_PER_BLOCK * BLOCKS_PER_PACKET);
 
+/** Special Definitions for VLS128 support **/
+// These are used to detect which bank of 32 lasers is in this block
+static const uint16_t VLS128_BANK_1 = 0xeeff;
+static const uint16_t VLS128_BANK_2 = 0xddff;
+static const uint16_t VLS128_BANK_3 = 0xccff;
+static const uint16_t VLS128_BANK_4 = 0xbbff;
+
+static const float  VLS128_CHANNEL_TDURATION  =  2.665f;  // [µs] Channels corresponds to one laser firing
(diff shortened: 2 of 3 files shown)


velodyne_driver/src/driver/driver.cc
```diff
@@ -62,7 +62,12 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
   private_nh.param("model", config_.model, std::string("64E"));
   double packet_rate;                   // packet frequency (Hz)
   std::string model_full_name;
-  if ((config_.model == "64E_S2") || 
+  if ((config_.model == "VLS128") )
+  {
+    packet_rate = 12507;     //https://github.com/velodynevls/velodyne_vls/blob/master/velodyne_driver/src/driver/driver.cc
+    model_full_name = std::string("HDL-") + config_.model;
```
JWhitleyWork, line 68: Shouldn't this be `VLS-` or just pass in the exact `config._model` value since you're expecting the VLS part to already be there?

velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
```diff
@@ -109,6 +109,19 @@ static const int BLOCKS_PER_PACKET = 12;
...
+/** Special Definitions for VLS128 support **/
+// These are used to detect which bank of 32 lasers is in this block
+static const uint16_t VLS128_BANK_1 = 0xeeff;
+static const uint16_t VLS128_BANK_2 = 0xddff;
+static const uint16_t VLS128_BANK_3 = 0xccff;
+static const uint16_t VLS128_BANK_4 = 0xbbff;
+
+static const float  VLS128_CHANNEL_TDURATION  =  2.665f;  // [µs] Channels corresponds to one laser firing
+static const float  VLS128_SEQ_TDURATION      =  53.3f;   // [µs] Sequence is a set of laser firings including recharging
+static const float    VLP32_DISTANCE_RESOLUTION   =    0.004f;  // [m]
```
JWhitleyWork, line 121: Shouldn't this be `VLS128_DISTANCE_RESOLUTION`?

velodyne_pointcloud/include/velodyne_pointcloud/rawdata.h
```diff
@@ -209,6 +225,16 @@ class RawData
...
+
+  void unpack_vls128(const velodyne_msgs::VelodynePacket &pkt, DataContainerBase& data,
+                     const ros::Time& scan_start_time);
+
+  /** in-line test whether a point is in range */
+  bool pointInRange(float range)
+  {
+    return (range >= config_.min_range
+            && range <= config_.max_range);
+  }
```
JWhitleyWork, line 237: You might as well give this the `inline` keyword.

velodyne_pointcloud/src/lib/rawdata.cc
```diff
@@ -196,7 +219,14 @@ inline float SQR(float val) { return val*val; }
...
   {
-    private_nh.param("model", config_.model, std::string("64E"));
+
+    if (!private_nh.getParam("model", config_.model))
+    {
+      config_.model =  std::string("64E");
+      ROS_ERROR("No Velodyne Sensor Model specified using default %s!", config_.model.c_str());
+
+    }
+//    private_nh.param("model", config_.model, std::string("64E"));
```
JWhitleyWork, line 229: Please remove commented-out code. We can use `git` to look back at what was here if necessary.

velodyne_pointcloud/src/lib/rawdata.cc
```diff
@@ -226,6 +256,11 @@ inline float SQR(float val) { return val*val; }
       cos_rot_table_[rot_index] = cosf(rotation);
       sin_rot_table_[rot_index] = sinf(rotation);
     }
+
+    for (uint8_t i = 0; i < 16; i++) {
+      vls_128_laser_azimuth_cache[i] = (VLS128_CHANNEL_TDURATION / VLS128_SEQ_TDURATION) * (i + i / 8);
+    }
```
JWhitleyWork, line 262: Please only calculate this if `config_.model == "VLS128"`.

velodyne_pointcloud/src/lib/rawdata.cc
```diff
@@ -255,6 +290,11 @@ inline float SQR(float val) { return val*val; }
...
```
JWhitleyWork, line 296: Can you please refactor `RawData::setup` and `RawData::setupOffline` so they call a shared function? There is a lot of duplicated code between these two.

velodyne_pointcloud/src/lib/rawdata.cc
```diff
@@ -269,6 +309,12 @@ inline float SQR(float val) { return val*val; }
     using velodyne_pointcloud::LaserCorrection;
     ROS_DEBUG_STREAM("Received packet, time: " << pkt.stamp);
 
+    /** special parsing for the VLS128 **/
+    if (pkt.data[1205] == 161 || pkt.data[1205] == 99) { // VLS 128
```
JWhitleyWork, line 313: What do these values mean? Is there somewhere they are documented?

velodyne_pointcloud/src/lib/rawdata.cc
```diff
@@ -431,6 +477,154 @@ inline float SQR(float val) { return val*val; }
...
+          //point.ring = corrections.laser_ring;
+          //point.x = xy_distance * cos_rot_angle;    // velodyne y
+          //point.y = -(xy_distance * sin_rot_angle); // velodyne x
+          //point.z = distance * sin_vert_angle;      // velodyne z
+
+          // Intensity extraction
+          //point.intensity = current_block.data[k + 2];
+
+          //pc.points.push_back(point);
+          //++pc.width;
```
JWhitleyWork, line 612: Please remove all commented-out code.

JWhitleyWork (changes requested):
See comments.

velodyne_pointcloud/src/lib/rawdata.cc
UniBwTAS: According to the manual the Byte in the position 1205 of the packet  corresponds to the Product ID. In the case of the VLS-128 the ID is 161. I can not find the 99 in the manual so I will remove it from the condition and make the 161 a constant in rawdata.h

JWhitleyWork: That's fine. Can you please add a comment in the code explaining this when you make the change?

velodyne_driver/src/driver/driver.cc
```diff
@@ -62,7 +62,12 @@ VelodyneDriver::VelodyneDriver(ros::NodeHandle node,
   private_nh.param("model", config_.model, std::string("64E"));
   double packet_rate;                   // packet frequency (Hz)
   std::string model_full_name;
-  if ((config_.model == "64E_S2") || 
+  if ((config_.model == "VLS128") )
+  {
+    packet_rate = 12507;     //https://github.com/velodynevls/velodyne_vls/blob/master/velodyne_driver/src/driver/driver.cc
```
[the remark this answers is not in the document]
UniBwTAS: We took this from the repository, but are currently evaluating it because  with the information of the manual we get only half of this value. We have already made some measurements using the first value and they look good but we will test the other value in the next couple of weeks.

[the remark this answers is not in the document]
JWhitleyWork: I would guess that this value is for when dual-return mode is active. By default, most Velodyne sensors are configured with either "strongest return" or "closest return" which provides only half the data. Not all sensors provide dual-return mode and this driver is currently not set up to support it so I recommend setting this to half the current size if that is what you're seeing from your sensor.

velodyne_pointcloud/src/lib/rawdata.cc
JWhitleyWork: Thanks for this! Much improved!

velodyne_driver/src/driver/driver.cc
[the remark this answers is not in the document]
UniBwTAS: Yea, that may be true. We think the value was calculated using wireshark or something similar, but provably they forgot to account for the positioning messages that the sensor sends. I have replaced it by the value calculated in the manual.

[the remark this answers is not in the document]
JWhitleyWork: LGTM!

JWhitleyWork (approved):

PR #378, opened 2020-12-18
Extended pointcloud

UniBwTAS:
The configuration of the lasers in the new velodyne Alpha prime allows an easy construction of a ordered point cloud resembling a depth image. The Point Cloud also contains some more information about each point that can be useful for segmentation algorithms.


JWhitleyWork:
@UniBwTAS This Pull Request is adding a large number of features (different point cloud types, new parameters, using ROS messages for Ethernet packets, etc.) in a single Pull Request and, for that reason, it is not easily review-able. Please open separate Pull Requests for each feature you wish to add to the driver.

PR #382, opened 2021-01-14
Add VLS128 launch and calibration file

wep21:
ROS1 driver has already supported VLS128 AP, but there are no calibration file and launch file, so I added them.


JWhitleyWork:
Thank you!

PR #384, opened 2021-01-19
PCAP timestamps & PCAP+GPS timestamps

HMellor:
This pull request continues on the work of #58 to add the ability to timestamp the published topics using the timestamp in the PCAP headers. I have also added the GPS time correction from #222 so that the PCAP ethernet header time can be corrected using the GPS time inside the Velodyne packet.

It handles the following combinations of the time parameters:
- `gps_time == false` and `pcap_time == false`: Sets time to `ros::Time::now()`
- `gps_time == false` and `pcap_time == true`: Sets time to the time in the PCAP ethernet headers
- `gps_time == true` and `pcap_time == false`: Sets time to `ros::Time::now()`, because this is an invalid combination of parameters
- `gps_time == true` and `pcap_time == true`: Sets time to the time in the PCAP ethernet headers and then uses the function from #222 to adjust it to GPS time

Looking at the respective info printouts for the PCAP and the bag (pcap_time and gps_time enabled rather than realtime headers), the timestamping appears to be functioning correctly:
```
user@pc:~$ rosbag info survey.bag 
...
start:       Nov 30 2020 15:18:13.10 (1606749493.10)
end:         Nov 30 2020 15:23:05.19 (1606749785.19)
...

user@pc:~$ capinfos survey.pcap 
...
First packet time:   2020-11-30 15:18:12.890587
Last packet time:    2020-11-30 15:23:05.169158
...
```

### DIFF
--- a/velodyne_driver/include/velodyne_driver/input.h
+++ b/velodyne_driver/include/velodyne_driver/input.h
@@ -139,6 +139,7 @@ private:
   bpf_program pcap_packet_filter_;
   char errbuf_[PCAP_ERRBUF_SIZE];
   bool empty_;
+  bool pcap_time_;
   bool read_once_;
   bool read_fast_;
   double repeat_delay_;
--- a/velodyne_driver/include/velodyne_driver/time_conversion.hpp
+++ b/velodyne_driver/include/velodyne_driver/time_conversion.hpp
@@ -59,7 +59,7 @@ ros::Time resolveHourAmbiguity(const ros::Time &stamp, const ros::Time &nominal_
     return retval;
 }
 
-ros::Time rosTimeFromGpsTimestamp(const uint8_t * const data) {
+ros::Time rosTimeFromGpsTimestamp(const uint8_t * const data, const struct pcap_pkthdr *header = NULL) {
     const int HOUR_TO_SEC = 3600;
     // time for each packet is a 4 byte uint
     // It is the number of microseconds from the top of the hour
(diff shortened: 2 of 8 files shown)


HMellor:
@JWhitleyWork would you mind taking a look at this since you oversaw #58?

JWhitleyWork:
@HMellor Sure. However, I disagree with the following assumption:

> `gps_time == true` and `pcap_time == false`: Sets time to ros::Time::now(), because this is an invalid combination of parameters

Instead, I would assume that this would indicate that the user wants to use the GPS time indicated in the packet contents and ignore the packet header times in the PCAP data. See how this is handled with real-time data [here](https://github.com/ros-drivers/velodyne/blob/f235ac6b0d1728e97de552f386b412f5fa5a092d/velodyne_driver/src/lib/input.cc#L234).

HMellor:
@JWhitleyWork Ah yes of course, my mistake. I was only thinking of the post process applications and forgot about the original intent of the `gps_time` parameter. I'll commit an update for that now.

HMellor:
Thanks @JWhitleyWork 
> Instead, I would assume that this would indicate that the user wants to use the GPS time indicated in the packet contents and ignore the packet header times in the PCAP data. See how this is handled with real-time data here.

The issue with that is that the real-time `gps_time == true` implementation doesn't actually extract the whole time from the PCAP data, it takes the hour from `ros::Time::now()` and adds the GPS microsecond from PCAP data, as can be seen [here](https://github.com/ros-drivers/velodyne/blob/f235ac6b0d1728e97de552f386b412f5fa5a092d/velodyne_driver/include/velodyne_driver/time_conversion.hpp#L70). 

Therefore, the functionality you describe is covered by the fourth combination from my original list, which uses the same technique to take the hour from the packet header and add the GPS microsecond using the PCAP data:
> * `gps_time == true` and `pcap_time == true`: Sets time to the time in the PCAP ethernet headers and then uses the function from #222 to adjust it to GPS time

This led me to conclude that, when using the `InputPCAP` class, setting `gps_time == true` and `pcap_time == false` is a degenerate case where you would be adding an old GPS microsecond to the current hour.

I see 2 alternatives for handling this when `pcap` is specified:
* Have `gps_time == true` and `pcap_time == false` throw an error that explains the degenerate case
* Have `gps_time == true` override the `pcap_time` parameter and automatically get the hour from the packet header and then add the GPS microsecond from the PCAP data

Which of these options do you prefer?

JWhitleyWork:
Ah, thanks for the explanation. This actually leads to something we should be using but aren't in the driver - the telemetry data on UDP port 8308. This contains the full GPRMC message from the GPS time source which includes the hour. I'll add an issue for this. For the time being, Here's my preference:

- `gps_time == true` (regardless of `pcap_time` value) - use UDP header for hour and modify with GPS time
- `gps_time == false` and `pcap_time == true` - use UDP header time
- `gps_time == false` and `pcap_time == false` - use `ros::Time::now()`

HMellor:
@JWhitleyWork I've pushed the change so that the arguments operate in the way you said you preferred, when you have the chance could please review and merge if you see no issues. Thanks

JWhitleyWork (approved):

PR #391, opened 2021-02-05
Fix accidental integer division in VLP-16 intensity calculation

valgur:


JWhitleyWork (approved):

JWhitleyWork:
Thanks for the fix! Can you please back-port this to the `melodic` branch?

PR #427, opened 2021-09-21
reuse Velodyne UDP port

AmetistDrake:
The Boston Dynamics Spot Core with the EAP (Enhanced Autonomy Package) version also binds to UDP port 2368 in its velodyne service, and without this update the ROS velodyne driver cannot bind to the port. It gives "Bind address already in use" warning and a "Cannot poll device" error. The reason is that, without setting the socket binding option to reuseable, it  locks away the UDP port from every other concurrent connection.


JWhitleyWork:
Close/open to kick CI.

JWhitleyWork:
CI failures are due only to CI server limitations. Approved.

JWhitleyWork (approved):

JWhitleyWork:
@AmetistDrake Would you mind forward-porting this to the `ros2` branch as well?

PR #428, opened 2021-10-11
fix timing bug with offline use

nickcharron:
Currently when using the offline setup, it does not build the per-point timing offsets which means when unpacking the packets in the offline mode, you cannot timestamp each point with its packet time.
This PR does the following:

    - call buildTimings from setupOffline function
    - add model to setupOffline interface
    - set config model param which is needed by buildTimings


xiaotaw:
Why this PR is not merged? It really helps.

pedroexenberger:
Couldn't this new method be exposed alongside the previous `setupOffline()`? Removing the previous method signature changes the interface and requires code changes after the library is upgraded.

For people who do not need this feature and, e.g., use docker to build and run some dependent code this causes unnecessary code adjustments.

PR #430, opened 2021-11-10
Replace deprecated argument names in launch

kquigley29:
Some of the argument names are deprecated in Foxy and do not work in Galactic. Eg. `node_executable` should now be just `executable`. After some testing with a VLP-16 on ROS2 Galactic the updated launch files worked as expected and the nodes were launched.


clalancette (approved):
Looks good to me, thanks for the fixes!

PR #434, opened 2021-12-26
Sync master

wep21:
Note:
1. https://github.com/ros-drivers/velodyne/commit/5206c95148c02abad71854d76adb52ec8c85f569
2. https://github.com/ros-drivers/velodyne/commit/3f11fab5f9a8d0774e566d63f1e6bbe9d7519341 -> drop. A part of this commit is already fixed in `ros2`. Another part shoud be fixed after parameter callback is implemented.
3. https://github.com/ros-drivers/velodyne/commit/0fc1aebbceef4ae8ceaf233267589e669fddef06 -> drop. already removed.
4. https://github.com/ros-drivers/velodyne/commit/0737cf077537ee7ddfa1d7771d22f759277ed74b
5. https://github.com/ros-drivers/velodyne/commit/96b39056bc16c5d3be9d32a709d4cd824189ef40 -> drop. already fixed.
6. https://github.com/ros-drivers/velodyne/commit/d3f85646503aa72a70aafab6ac757c36aa3ff1c2
7. https://github.com/ros-drivers/velodyne/commit/f028b5972c8a179b951117a42eac69e0505a665d -> drop. unnecessary for `ros2`
8. https://github.com/ros-drivers/velodyne/commit/720829573076bf7d7edd4b1de344924fad909049 -> drop. A poll thread is implemented in `ros2`.

PR #437, opened 2022-01-04
bug fix for row step

wep21:
ros2 porting for #250
a part of #434

### DIFF
--- a/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.hpp
+++ b/velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.hpp
@@ -83,7 +83,7 @@ public:
     cloud.width = config_.init_width;
     cloud.height = config_.init_height;
     cloud.is_dense = static_cast<uint8_t>(config_.is_dense);
-    cloud.row_step = cloud.width * cloud.point_step;
+    cloud.row_step = init_width * cloud.point_step;
   }
 
   virtual ~DataContainerBase() {}


wep21:
cc @clalancette @JWhitleyWork

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.hpp
```diff
@@ -83,7 +83,7 @@ class DataContainerBase
     cloud.width = config_.init_width;
     cloud.height = config_.init_height;
     cloud.is_dense = static_cast<uint8_t>(config_.is_dense);
-    cloud.row_step = cloud.width * cloud.point_step;
+    cloud.row_step = init_width * cloud.point_step;
```
clalancette, line 86: I don't think this applies to the `ros2` branch.

`init_width` is passed into the constructor.  We then use `init_width` to initialize `config_.init_width` in line 63-64 above.  We then initialize `cloud.width` from `config_.init_width` in line 83.  So at this point, `cloud.width` should always be the same as `init_width`, and thus I don't think this is necessary.

clalancette (changes requested):

velodyne_pointcloud/include/velodyne_pointcloud/datacontainerbase.hpp
wep21: You're right. I will drop this.

PR #290, opened 2019-10-02
Add per-point time field

kaarta-SHanna:
Adds a float field "time" indicating the per-point firing time (relative to the pointcloud's header).

PR #440, opened 2022-01-05
Add per point time field

wep21:
ros2 porting for #290
a part of #434

### DIFF
--- a/velodyne_pointcloud/config/VLP16-velodyne_convert_node-params.yaml
+++ b/velodyne_pointcloud/config/VLP16-velodyne_convert_node-params.yaml
@@ -1,6 +1,7 @@
 velodyne_convert_node:
     ros__parameters:
         calibration: VLP16db.yaml
+        model: VLP16
         min_range: 0.9
         max_range: 130.0
         view_direction: 0.0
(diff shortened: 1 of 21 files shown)


wep21:
cc @JWhitleyWork @clalancette

wep21:
@clalancette @JWhitleyWork friendly ping

wep21:
@clalancette @JWhitleyWork friendly ping

JWhitleyWork:
@wep21 Sorry for taking so long to get back to you. I believe this field should be added in a new point cloud container - something like `PointXYZIRT`. If it is added to the existing container, it will cause issues with data that were recorded using the existing point structure.

wep21:
@JWhitleyWork I will rename `PointXYZIR` `PointXYZIRT` in https://github.com/ros-drivers/velodyne/pull/434/commits/749d3bb5c9a63aeae091aedc7301b3dc12efcd65. Is it enough for your suggestion?

JWhitleyWork:
That should be fine. Please make sure to rename everywhere in the code.

wep21:
@JWhitleyWork rename at https://github.com/ros-drivers/velodyne/pull/440/commits/3d529c523f17b538c8e0a07cd4039059960f57b2

JWhitleyWork:
FYI - I'm not a fan of the fact that this PR introduces two changes - one for the time field and one for making `model` a new, required parameter. Please separate these into separate PRs in the future. Since this is going to the `ros2` branch, which is basically targeted at Rolling at this point, I'm willing to let it slide for now.

JWhitleyWork (approved):

PR #442, opened 2022-01-06
reuse Velodyne UDP port (#427)

wep21:
ros2 porting for #427
a part of #434

### DIFF
--- a/velodyne_driver/src/lib/input.cpp
+++ b/velodyne_driver/src/lib/input.cpp
@@ -129,7 +129,7 @@ InputSocket::InputSocket(
   // compatibility with Spot Core EAP, reuse port 2368
   int val = 1;
   if (setsockopt(sockfd_, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val)) == -1) {
-    perror("socketopt");
+    RCLCPP_ERROR(private_nh->get_logger(), "Error setting REUSEADDR: %s", ::strerror(errno));
     return;
   }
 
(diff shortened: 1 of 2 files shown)


wep21:
cc @clalancette @JWhitleyWork

velodyne_driver/src/lib/input.cpp
```diff
@@ -126,6 +126,13 @@ InputSocket::InputSocket(
   my_addr.sin_port = htons(port);          // port in network byte order
   my_addr.sin_addr.s_addr = INADDR_ANY;    // automatically fill in my IP
 
+  // compatibility with Spot Core EAP, reuse port 2368
+  int val = 1;
+  if (setsockopt(sockfd_, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(val)) == -1) {
+    perror("socketopt");
```
clalancette, line 132: Instead of using perror here, I think we should use `RCLCPP_ERROR` like we do elsewhere.  So maybe something more like:

```suggestion
    RCLCPP_ERROR(private_nh->get_logger(), "Error setting REUSEADDR: %s", ::strerror(errno));
```

clalancette (changes requested):
Minor change, then I'm good with this.

velodyne_driver/src/lib/input.cpp
wep21: @clalancette accepted your suggestion at https://github.com/ros-drivers/velodyne/pull/442/commits/aa8220969f9887d8f219198cfadf057f751f0282.

clalancette (approved):

PR #446, opened 2022-02-02
Velodyne pcl (#335)

wep21:
ros2 porting of #335
A part of https://github.com/ros-drivers/velodyne/pull/434

### DIFF
--- a/velodyne_pcl/CMakeLists.txt
+++ b/velodyne_pcl/CMakeLists.txt
@@ -1,0 +1,35 @@
+cmake_minimum_required(VERSION 3.5)
+project(velodyne_pcl)
+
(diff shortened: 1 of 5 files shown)


wep21:
cc @clalancette @JWhitleyWork

JWhitleyWork:
@wep21 Myself and @spuetz decided to leave this out of the ROS2 version as the PCL dependency is very heavy and that also depends on *all of Boost*, which is also very heavy. Is there a particular reason you need this to be in the driver?

wep21:
@JWhitleyWork There is no particular reason and this is a utility function for external use, so it is possible to proceed to sync master #434 without this PR.

Issue #445, opened 2022-02-01
Noetic PCL Build warning (-Wdeprecated-declarations)


icolwell-as:
**Please complete the following information:**
 - Ubuntu 20.04
 - Noetic
 - Installed from apt
 - Version: 1.6.1-1focal.20220107.012158

**Describe the bug**
When including `velodyne_pcl/point_types.h` for use in a ROS package, the following error is produced when compiling:

```
                 from /home/ian/autoware.ai/src/core_perception/points_downsampler/nodes/ring_filter/ring_filter.cpp:17:
/opt/ros/noetic/include/velodyne_pcl/point_types.h:66:36: warning: ‘using uint16_t = uint16_t’ is deprecated: use std::uint16_t instead of pcl::uint16_t [-Wdeprecated-declarations]
   66 |                                   (uint16_t, ring, ring)
      |                                    ^~~~~~~~
In file included from /usr/include/pcl-1.10/pcl/point_types.h:42,
                 from /opt/ros/noetic/include/velodyne_pcl/point_types.h:46,
                 from /home/ian/autoware.ai/src/core_perception/points_downsampler/nodes/ring_filter/ring_filter.cpp:20:
/usr/include/pcl-1.10/pcl/pcl_macros.h:94:9: note: declared here
   94 |   using uint16_t [[deprecated("use std::uint16_t instead of pcl::uint16_t")]] = std::uint16_t;
      |         ^~~~~~~~

```

**To Reproduce**

Include the header and compile it.

**Expected behavior**
No deprecation warnings.

**Additional context**
I think the problem might actually be coming from the specific version of PCL installed.

JWhitleyWork:
@icolwell-as Would you mind submitting a PR to resolve?

icolwell-as:
@JWhitleyWork, I had actually intended to do that, but then got a bit lost trying to find the real root cause. I could open a PR that would resolve it, but it might be better resolved some other way. 

Perhaps you can help clear up some of my confusion: When it comes to the version of PCL that is used with certain ROS distros (noetic for example), it is pretty much dictated by the version of PCL that is distributed from Ubuntu's repos right? in the case of Ubuntu 20.04 this appears to be 1.10.0 when I run `dpkg -l libpcl-dev`.

This version of PCL will pretty much remain the same for the lifetime of ROS Noetic right? we wouldn't expect it to get updated due to Ubuntu's "stability" approach to their releases?

JWhitleyWork:
[...]

PR #449, opened 2022-02-07
Use std::uint16_t to reduce build warnings from PCL

icolwell-as:
Resolves https://github.com/ros-drivers/velodyne/issues/445.

Not really a necessary change, but it gets rid of deprecated build warnings, and is more explicit about the fact that it isn't using PCL's (now deprecated) definition of `uint16_t`.


JWhitleyWork (approved):
LGTM.

PR #453, opened 2022-03-05
feat: unify tf frame parameters between transform and cloud nodes

wep21:
a part of #434
ros2 porting of #344

* Unify tf frame parameters between transform and cloud nodes

* Reformat code slightly to avoid roslint errors

* Use boost bind instead of lambdas to add support for kinetic

* Fix some search and replace errors in comments

* Add fixed_frame and target_frame params back

* Use more common ROS terminology instead of 'htm'

* Undo renaming of publisher and subscriber

* Just use tf listener when necessary

* Migrate package to tf2

* Explicitly store sensor frame in a dedicated variable to make code easier to read

* Reformat code slightly to avoid roslint errors

* Avoid unnecessary transforms when sensor_frame == target_frame

* Fix comments

Co-authored-by: anre <andreas.reich@unibw.de>


wep21:
FYI @JWhitleyWork @clalancette

wep21:
@JWhitleyWork @clalancette friendly ping

wep21:
@JWhitleyWork @clalancette friendly ping

PR #481, opened 2022-10-25
Feature/timestamp from phaselock angle

PhilippSchmaelzle:
Adds the possibility to set the point cloud time stamp to the stamp of the package received when passing the phase_lock angle.
It is implemented similar to the cut_off angle and makes probably only sense to use both together. 

I tested it live with an VLP32-c

Addresses also #242

### DIFF
--- a/velodyne_driver/include/velodyne_driver/driver.h
+++ b/velodyne_driver/include/velodyne_driver/driver.h
@@ -74,15 +74,18 @@ private:
     int    npackets;                 // number of packets to collect
     double rpm;                      // device rotation rate (RPMs)
     int cut_angle;                   // cutting angle in 1/100°
+    int phase_lock_angle;            // configured phase lock angle
     double time_offset;              // time in seconds added to each velodyne time stamp
     bool enabled;                    // polling is enabled
     bool timestamp_first_packet;
+    bool timestamp_phase_lock;
   }
   config_;
 
   boost::shared_ptr<Input> input_;
   ros::Publisher output_;
   int last_azimuth_;
+  int phase_lock_angle_;
 
   /* diagnostics updater */
   ros::Timer diag_timer_;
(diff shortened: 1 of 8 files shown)


velodyne_driver/src/driver/driver.cc
```diff
@@ -237,6 +264,14 @@ bool VelodyneDriver::poll(void)
       	 last_azimuth_ = azimuth;
       	 continue;
       }
+
+      if((last_azimuth_ < config_.timestamp_angle && config_.timestamp_angle <= azimuth)
+         || ( config_.timestamp_angle <= azimuth && azimuth < last_azimuth_)
+         || (azimuth < last_azimuth_ && last_azimuth_ < config_.timestamp_angle))
```
mpitropov, line 270: The logic in this if statement is the same logic as line 275 and should really be placed in its own function to be used on both lines
```
if (angleIsBetween(last_azimuth_, azimuth, config_.timestamp_angle))
```

Within the function, I think you need a comment just explaining the logic since it was not immediately obvious to me.
```
// Case 1: 0° < last azimuth < angle < azimuth < 360°
// Case 2: last azimuth < 360° < 0° < angle < azimuth
// Case 3: last azimuth < angle < 360° < 0° < azimuth
```

I'm not even completely sure what is going on since there are other cases listed in the issue here but I don't know how those show up https://github.com/ros-drivers/velodyne/issues/174

mpitropov (commented):
I was checking GitHub and noticed my issue was mentioned in this MR. I came up with a hack back then but this looks like a proper way to add the feature. I have some comments to reduce redundant code and make it a bit easier to understand for future people who are new to the codebase.

JWhitleyWork:
Please address comments by @mpitropov.

velodyne_driver/include/velodyne_driver/driver.h
```diff
@@ -74,6 +74,7 @@ class VelodyneDriver
     int    npackets;                 // number of packets to collect
     double rpm;                      // device rotation rate (RPMs)
     int cut_angle;                   // cutting angle in 1/100°
+    int timestamp_angle;            // configured phase lock angle
```
JWhitleyWork, line 77: Could you please include a description somewhere of what this actually means/does? It's definitely unclear to me.

PhilippSchmaelzle:
Thanks for the feedback!
I will try to address it soon. I do not find the time for it at the moment.

PR #497, opened 2022-12-17
Disable cut_angle by default.

JWhitleyWork:
Prior to the `cut_angle` feature being added, the default behavior was to consider a complete rotation to be a set number of packets. This feature being added and enabled by default changed the default behavior subtley and caused undesired effects. This PR reverts that change but leaves the `cut_angle` feature available for those who choose to use it.


JWhitleyWork:
@cjffly Please confirm that this change resolves your issue.

PR #509, opened 2023-04-17
add publishing of laser beam visualization markers

mantkiew:
It is useful to be able to see the lidar beams in RViz. This PR adds publishing of RViz markers for all lidar beams according to the lidars calibration parameters and range. We have used this functionality to validate dual lidar design on University of Waterloo's UW Moose:

lidar-frames-10-deg-close

ground_lidar-10-deg


JWhitleyWork:
While I think this could be useful, having it enabled by default is definitely not desirable. Please add a parameter to enable the feature, have the parameter default to `False`, and wrap the creation of the markers in an `if`.

PR #531, opened 2023-09-25
Fix time_offset units

nachovizzo:
`time_offset` is specified everywhere to be in `seconds`. Probably something that was ported from the ROS 1 implementation. In the changed line, the summation was done in `nanoseconds`, and therefore, the maximum time offset one could add was just `1.0` nanoseconds

Done in Collaboration with @doisyg


nachovizzo:
Closing this due to lack of activity :) feel free to reopen if needed

PR #535, opened 2024-01-08
Fix exports

g-kurz:
This pull request fixes some exports that are needed when building applications or libraries that link to the velodyne driver.


clalancette (approved):
I ended up retargeting this to `ros2` (which is what we use for everything right now), and also updating the exports.

Thanks for the contribution!

PR #537, opened 2024-01-28
fix qos to system default  to Unlocking the potential of Fast DDS middleware

anscipione:


clalancette:
While I understand the desire to do this, I think that this is actually a mistake.

The biggest problem here is the SystemDefaultsQoS ends up setting the queue depth to 1 (in all of our current implementations), which mean that it is going to be much more likely to drop data.  So I'm going to close this one out for now, but do feel free to reopen if you'd like to discuss more.

PR #542, opened 2024-10-01
feat(config): make parameter 'enabled' dynamic

reinzor:
So that it can be used no run-time for enabling / disabling the driver.


clalancette:
Closing in favor of #548 (which takes this commit, and fixes uncrustify).

PR #544, opened 2024-10-29
fix(build): add missing `eigen` dependency

### LINKED ISSUE #543: Add `eigen` as a dependency to `velodyne_pointcloud`
> This removes a dependency of `velodyne_pointcloud` that is still necessary in ROS 2 (see https://github.com/lopsided98/nix-ros-overlay/issues/511#issuecomment-2441025459 for an example of an issue this causes) 

 _Originally posted by @quantum9Innovation in [cd47706](https://github.com/ros-drivers/velodyne/commit/cd47706f8a57c46477b536e9749f372c7e75f40b#r148505456)_

quantum9Innovation:
close #543

### DIFF
--- a/velodyne_pointcloud/package.xml
+++ b/velodyne_pointcloud/package.xml
@@ -30,6 +30,7 @@
   <depend>tf2_ros</depend>
   <depend>velodyne_msgs</depend>
   <depend>yaml-cpp</depend>
+  <depend>eigen</depend>
 
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_auto</test_depend>


velodyne_pointcloud/package.xml
```diff
@@ -30,6 +30,7 @@
   <depend>tf2_ros</depend>
   <depend>velodyne_msgs</depend>
   <depend>yaml-cpp</depend>
+  <depend>eigen</depend>
```
wentasah, line 33:
```suggestion
  <depend>eigen</depend>
  
  <build_depend>pkg-config</build_depend>
```

I think that `<build_depend>pkg-config</build_depend>` should be added too. Without it, I'm getting this error:
```
...
> -- Found tf2_ros: 0.25.6 (/nix/store/6pxj2y6qkl5wlfmhjypkb9j4j95pmij9-ros-humble-tf2-ros-0.25.6-r1/share/tf2_ros/cmake)
> -- Found velodyne_msgs: 2.4.0 (/nix/store/xm5qphp0mqs8dbq14328g25ph8fys0im-ros-rolling-velodyne-msgs-2.4.0/share/velodyne_msgs/cmake)
> CMake Error at /nix/store/jacf2kn4dfj99c3ywbvfyg6w53xvxsfl-cmake-3.27.8/share/cmake-3.27/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
>   Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
> Call Stack (most recent call first):
>   /nix/store/jacf2kn4dfj99c3ywbvfyg6w53xvxsfl-cmake-3.27.8/share/cmake-3.27/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
>   /nix/store/jacf2kn4dfj99c3ywbvfyg6w53xvxsfl-cmake-3.27.8/share/cmake-3.27/Modules/FindPkgConfig.cmake:99 (find_package_handle_standard_args)
>   CMakeLists.txt:29 (find_package)
```

wentasah (changes requested):

velodyne_pointcloud/package.xml
clalancette: Yes, you are correct that we also need `pkg-config` here.

That said, this should be temporary, as nowadays we should be able to find yaml-cpp with pure CMake.  But that will be a separate PR.

clalancette:
Given that I wasn't able to push here, I ended up opening a new PR in #545 to add in the pkg-config dependency.  Thanks for this, I'm going to close this out and merge that one instead.

PR #556, opened 2025-01-21
[ros2] Fix parameter loading in launch files (VLP16, VLP32C, VLS128)Fix launch file params

JVALPASS:
This PR addresses a bug where the Velodyne driver node was incorrectly passing
the entire YAML path to the `parameters` argument rather than loading the
dictionary of parameters.

Changes:
* Updated `velodyne_driver_node-VLP16-launch.py` to load YAML using `yaml.safe_load`.
* Updated `velodyne_driver_node-VLP32C-launch.py` similarly.
* Updated `velodyne_driver_node-VLS128-launch.py` similarly.
* Verified that parameters are now loaded correctly under `velodyne_driver_node.ros__parameters`.

Tests:
* Launched each file in a ROS2 workspace and confirmed correct parameter loading.

### DIFF
--- a/velodyne_driver/launch/velodyne_driver_node-VLP16-launch.py
+++ b/velodyne_driver/launch/velodyne_driver_node-VLP16-launch.py
@@ -33,7 +33,7 @@
 """Launch the velodyne driver node with default configuration."""
 
 import os
-
+import yaml 
 import ament_index_python.packages
 import launch
 import launch_ros.actions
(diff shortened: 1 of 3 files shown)


benjaminholdennearearth:
The `Node` action in `launch_ros` supports input as a list of paths to yaml files. No need to load it yourself in the launch file.
https://github.com/ros2/launch_ros/blob/humble/launch_ros/launch_ros/actions/node.py#L163-L167

JVALPASS:
Thank you for your feedback! I understand, and you’re absolutely right. I’ll close this PR since it's not needed.

Issue #557, opened 2025-04-22
ROS 2 Velodyne Transform Node does not publish /velodyne_points at 10 Hz (Jazzy)


ming0071:
🛠 System Information

- OS: Ubuntu 24.04
- ROS 2 Version: Jazzy
- Installation Method: Cloned from ros2 branch

📌 Issue Summary
When running `velodyne_pointcloud::Transform` on ROS 2 Jazzy, the node correctly subscribes to `/velodyne_packets` and receives data at 10 Hz. However, the output topic `/velodyne_points` is either published at a very low frequency (sometimes < 1 Hz) or not published at all — despite seeing the point cloud appear in RViz.

In contrast, the ROS 1 version (using the same lidar and parameters) consistently publishes `/velodyne_points` at 10 Hz.

🔍 Observations

Adjusting the `min_range` parameter in ROS 2 has a significant effect:

Low values (e.g., `min_range: 0.1`) → `/velodyne_points`rarely or never publishes.

Higher values (e.g., `min_range: 2.0` or above) → publishing becomes stable.

It appears the node will only publish when valid points within the given `min_range` exist, suggesting there is some kind of filtering or accumulation threshold internally.

The `cut_angle` parameter was commented out and not affecting the issue.

📂 Configuration

`VLP16-velodyne_driver_node-params.yaml`:
```
velodyne_driver_node:
    ros__parameters:
        device_ip: 192.168.2.81
        gps_time: false
        # cut_angle: 0.0
        time_offset: 0.0
        enabled: true
        read_once: false
        read_fast: false
        repeat_delay: 0.0
        frame_id: velodyne
        model: VLP16
        rpm: 600.0
        port: 2368
        timestamp_first_packet: false

```

`VLP16-velodyne_transform_node-params.yaml`：
```
velodyne_transform_node:
    ros__parameters:
        calibration: VLP16db.yaml
        model: VLP16
        min_range: 2.0
        max_range: 130.0
        view_direction: 0.0
        fixed_frame: "base_link"
        target_frame: "velodyne"
        organize_cloud: false
```

![Image](https://github.com/user-attachments/assets/c374181c-812e-4494-998d-5befd23c2552)
![Image](https://github.com/user-attachments/assets/931ebee7-02a7-442f-bb9e-99d80b8d6644)
❓ Questions

Has the publish behavior of `velodyne_pointcloud::Transform` changed in ROS 2 compared to ROS 1?

[...]

PR #565, opened 2025-08-08
add lauch file using zenoh middleware for publishing high throughput pointcloud

justyx404:
With the middleware installed, this should able to solve the slow publishing rate #557  that was caused by the default DDS middleware, tested on `humble`


JWhitleyWork:
To be honest, I'm not a big fan of adding this since it makes a lot of assumptions about the user's environment (that they have Zenoh installed, that they aren't already running a router, etc.). This should be something left to the user. However, I would support a PR to add a note to the REAMDE about high-throughput lidar scans and Zenoh being a possible solution.

justyx404:
Will do that instead. Just want to save people's time if someone else ever run into the throughput bottleneck.

PR #567, opened 2025-10-20
include Zenoh info in README

justyx404:
Add the Zenoh middle ware notes as suggested.


JWhitleyWork:
I'm not real fond of the idea of promiting a specific middleware over others. The recommendation to use a different middleware from the default is fine but please remove the Zenoh-specific information.

justyx404:
Sure.

